PR #14, opened 2018-05-29
NNW-52 Add houseblock results screen and middleware - Display results…

steverendell:
… table


steverendell:
Now has new agreed data format (main activity + other events)

PR #19, opened 2018-06-04
NN-865 houseblock results again

steverendell:
Fix style problems, tests, webpack


steverendell:
Just the last commit, the rest is already in.

PR #22, opened 2018-06-06
NN-865 Houseblock results ordering

steverendell:


steverendell:
two separate commits here - ordering and spinner/error clearing

PR #25, opened 2018-06-11
fix activity display and renamed a label

JamieDraperUK:


JamieDraperUK:
Yeah it has been tested with dev data.

PR #38, opened 2018-06-19
NN-867, NN-838 Activity / Work Shop Searching, Display and print

steverendell:
Functionally complete I think but tests to follow


backend/controllers/activityList.js
```diff
@@ -1,26 +1,42 @@
...
   const viewModel = await getActivityList(req, res);
   res.json(viewModel);
 }));
 
-const getActivityList = (async (req, res) => {
-  req.query.usage = 'PROD';
-  const activities = await elite2Api.getActivityList(req, res).data;
+const getActivityList = async (req, res) => {
+  // once request is not used passed to downstream services we wont have to manipulate it in this dodgy way
+  switchDateFormat(req);
```
steverendell: needs tidying

src/ResultsActivity/ResultsActivity.js
```diff
@@ -0,0 +1,183 @@
...
+import moment from 'moment';
+import { Link } from "react-router-dom";
+
+class ResultsActivity extends Component {
+  displayBack () {
+    return (<div className="padding-top no-print"><Link id={`back_to_menu_link`} title="Back to selection screen link" className="link backlink" to="/whereaboutssearch" >
+      <img className="back-triangle" src="/images/BackTriangle.png" alt="" width="6" height="10"/> Back to selection screen</Link></div>);
+  }
+
+  getActivityName () {
```
steverendell: a.locationId is an integer, is there a better way to compare these?

backend/controllers/activityList.js
JamieDraperUK: Why not add start splitting request parameters out now? elite2api.getActivities(req,red,date) should be fairly straight forward

src/ResultsActivity/ResultsActivity.js
JamieDraperUK: You could do this.props.activities.filter(a => a.locationId  === Number(this.props.activity)

src/ResultsActivity/ResultsActivity.js
```diff
@@ -0,0 +1,183 @@
...
+        this.props.handleSave(this.props.history);
+      }}>Save changes</button>
+      {this.isToday(this.props.date) &&
+      <button id="printButton" className="button greyButton rightHandSide" type="button" onClick={() => {
+        this.props.handlePrint();
+      }}><img className="print-icon" src="/images/Printer_icon.png" height="23" width="20"/> Print list</button>
+      }
+    </div>);
+
+    let activityTitle;
```
JamieDraperUK: Might not be a massive deal but this is duplication from the ResultsHouseBlock

PR #50, opened 2018-07-06
Nn 968 Remove redundant cookies from prisonstaffhub.

bell-pa:


backend/validate-client-version.js
```diff
@@ -1,24 +0,0 @@
-const log = require('./log');
```
steverendell: is it ok deleting the version checker?

bell-pa: Yes.  The version checker is a problem: if the version changes the check forces a re-load of the app, so they will lose whatever the current state of the app is. Mike felt that because we want to be able to deploy new versions without taking anything out of service this was unacceptable.

PR #62, opened 2018-07-23
Nn 1003 whereabouts blank screen

JamieDraperUK:


JamieDraperUK:
The activity list and house block list screen both share the same reducer. The event data is stored in a houseblock.data. When navigating between pages the data isn't cleared and would result in one screen trying to process data that's in the wrong shape. The fix is to store data into it's own property.

PR #63, opened 2018-07-24
NN-1017 Middle tier for attendance update: pay and other

steverendell:


backend/controllers/updateAttendance.js
```diff
@@ -0,0 +1,32 @@
...
+    throw new Error("Offender or activity id missing");
+  }
+  const { offenderNo, activityId } = req.query;
+  await elite2Api.updateAttendance(req, { offenderNo, activityId }, res);
+  log.info(res.status, 'updateAttendance success');
+  if (req.body.eventOutcome === 'UNACAB' && req.body.performance === 'UNACCEPT') {
+    req.body = {
+      type: 'Negative Behaviour',
+      subType: 'IEP Warning',
+      occurrenceDateTime: moment().format('YYYY-MM-DDThh:mm'),
```
JamieDraperUK: This may be UTC, might be worth forcing it to local time.

backend/controllers/updateAttendance.js
```diff
@@ -0,0 +1,32 @@
...
+      text: 'Refused to attend activity / education.'
```
JamieDraperUK: You may have to tweak this depending on the other story. I think the case note is triggered by the same code, however, it requires a different comment.

PR #67, opened 2018-07-27
NN-1007-various layout tweaks

JamieDraperUK:


JamieDraperUK:
Might leave them in for now. They'll be trendy again by Monday :)

PR #78, opened 2018-08-03
Nn 1026 only one main activity

steverendell:
Id look at the commits separately as I cleaned up a lot of test warnings

### DIFF
--- a/backend/controllers/houseblockList.js
+++ b/backend/controllers/houseblockList.js
@@ -3,6 +3,15 @@ const log = require('../log');
 const switchDateFormat = require('../utils');
 const moment = require('moment');
 
+function addToOthers (currentRow, event) {
+  if (!currentRow.others) {
+    // set array of non-paid activities, appointments or visits
+    currentRow.others = [event];
+  } else {
+    currentRow.others.push(event);
+  }
+}
+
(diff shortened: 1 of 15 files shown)


backend/controllers/houseblockList.js
```diff
@@ -22,20 +31,29 @@ const getHouseblockListFactory = (elite2Api) => {
...
         } else {
-          currentRow.activity.push(event);
-          currentRow.activity.sort(function (a, b) {
-            return safeTimeCompare(a, b);
-          });
+          // Multiple activities - make paid the main one
+          if (!currentRow.activity.payRate && event.payRate) {
+            addToOthers(currentRow, currentRow.activity);
+            currentRow.activity = event;
+          } else if ((currentRow.activity.payRate && event.payRate) ||
```
JamieDraperUK: Could this just be an else? It seems to be true if null or not null.

JamieDraperUK: Actually, my comment isn't true. You checking if both are true or both are false. I wonder if there's a cleaner way to describe the meaning of the different combinations.

steverendell: possibly. Ill have a think and if so  commit it with the next ticket

PR #107, opened 2018-09-03
NN 1088 fix location list sorting

JamieDraperUK:


backend/utils.js
```diff
@@ -1,5 +1,25 @@
...
+var reN = /[^0-9]/g;
+
+function sortAlphaNum (aa, bb) {
+  const a = (aa || "");
+  const b = (bb || "");
+
+  const aAlpha = a.replace(reA, "");
+  const bAlpha = b.replace(reA, "");
+
+  if (aAlpha === bAlpha) {
```
steverendell: sometimes a number comes first , e.g. 5-a-side, or numbers and letters are interleaved 'Hb5a2 Classroom'.
Perhaps stick to a normal string sort unless the pattern matches 'just letters' then 'just numbers'

backend/tests/utils.test.js
```diff
@@ -0,0 +1,21 @@
+const { sortAlphaNum } = require('../utils');
+
+describe('sortAlphaNum', () => {
+  it('should sort list correctly', () => {
+    const data = [
+      'WORKSHOP 11',
```
steverendell: Maybe include a single digit, and other strings without digits, or a complex mixture?

PR #122, opened 2018-09-18
NN-1144 Court event status descriptions

JamieDraperUK:
**Expose court event status descriptions**


backend/tests/activityList.test.js
```diff
@@ -309,44 +317,6 @@ describe('Activity list controller', async () => {
       undefined
     ]);
   });
```
steverendell: Couple of tests being deleted .. are we losing coverage?

JamieDraperUK: No, I turned these into integration tests as I didn't duplicate test logic.

PR #128, opened 2018-09-21
NN-1166 Should be possible to scroll app  so that an overflowing Menu…

bell-pa:
… dropdown can be accessed.

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -49,7 +49,7 @@ body {
     display: flex;
     width: 100%;
     flex-direction: column;
-    overflow: hidden;
+    overflow: scroll;
   }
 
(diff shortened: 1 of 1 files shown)


src/index.scss
```diff
@@ -49,7 +49,7 @@ body {
     display: flex;
     width: 100%;
     flex-direction: column;
-    overflow: hidden;
+    overflow: scroll;
```
jsrobertson: I'm not sure it even needs `scroll`.  Think the `overflow` property could be removed altogether.

bell-pa: Good point. Thanks. I'll try that and see...

PR #188, opened 2018-11-16
NN-1417: Add in cookie session instead of manual hmpps cookie handling

petergphillips:


backend/contextProperties.js
```diff
@@ -1,32 +1,21 @@
...
- * Note that by convention the controller(s) and Middleware use the res.locals property as the request scoped context.
+ * Note that by convention the controller(s) and Middleware use the req.session property as the request scoped context.
  * From controllers down to clients, client interceptors etc the context object is called 'context'.
  */
 
 /* eslint no-param-reassign: 2 */
 const setTokens = (context, accessToken, refreshToken) => {
-  context.accessToken = accessToken
-  context.refreshToken = refreshToken
+  context.access_token = accessToken
```
petergphillips: had to change this as the names of the variables changed between the context and the cookie, so need to use the cookie values to keep consistent with other apps

backend/index.js
```diff
@@ -117,19 +114,24 @@ const controller = controllerFactory(
...
-  secure: config.app.production,
-})
+app.use(
+  cookieSession({
+    name: config.hmppsCookie.name,
+    domain: config.hmppsCookie.domain,
+    maxAge: config.hmppsCookie.expiryMinutes * 60 * 1000,
+    secure: config.app.production,
+    httpOnly: true,
+    signed: false,
```
petergphillips: `signed` should be `true`, but can't add it in until all apps using cookie session and we've added the key to the environments

src/ResultsActivity/ResultsActivity.test.js
```diff
@@ -158,6 +158,7 @@ describe('Offender activity list results component', () => {
         agencyId={PRISON}
         showPaymentReasonModal={jest.fn()}
         user={user}
+        resetErrorDispatch={jest.fn()}
```
petergphillips: fixing warnings in the build

backend/contextProperties.js
```diff
@@ -1,32 +1,21 @@
...
+  context.refresh_token = refreshToken
 }
 
-const hasTokens = context => {
-  if (!context) return false
-  return !(!context.accessToken || !context.refreshToken)
-}
+const hasTokens = context => !!(context && context.access_token && context.refresh_token)
```
JamieDraperUK: Might be more readable by replacing the double bang with Boolean(context && context.access_token && context.refresh_token)

petergphillips: Like the idea, will do

PR #204, opened 2018-11-29
Nn 1452 location filter

lbennett-moj:


src/GlobalSearch/GlobalSearchForm.js
```diff
@@ -1,43 +1,164 @@
...
+      handleSearchGenderFilterChange,
+      handleSearchLocationFilterChange,
+      clearFilters,
+    } = this.props
+
+    const toggleDetails = (event, clear) => {
+      event.preventDefault()
+      const { showFilters } = this.state
+      this.setState({
+        showFilters: !showFilters,
```
bell-pa: According to [the react docs](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) you should use a different version of setState if you're referring to the current state.
`this.setState((state) => { showFilters: !state.showFilters }`

src/GlobalSearch/GlobalSearchForm.js
```diff
@@ -1,43 +1,164 @@
...
+            aria-checked={showFilters}
+          >
+            <span className="govuk-details__summary-text">{showFilters ? 'Hide filters' : 'Show filters'}</span>
+          </summary>
+          <div className="govuk-details__text">
+            <div className="pure-u-md-1-4 padding-top padding-bottom-large">{locationSelect}</div>
+            <div className="pure-u-md-1-4 padding-top padding-bottom-large padding-left">{genderSelect}</div>
+            <div>
+              <a className="clear-filters link clickable" {...linkOnClick(clearFilters)}>
+                Clear filters
```
bell-pa: How about making the clear filters link  a component?  If you did that it would encapsulate the props that you're supplying using a function. Then there'd be no need to put a 'linkOnClick' function in utils.js.

PR #205, opened 2018-11-30
NN-1398 establishment roll back

jsrobertson:
Added <Page /> component that was created in omic-ui to help us standardise and create new pages easily. So it handles the following in one place:
- Breadcrumbs/back button
- Loading state
- Error state
- Page titles

### DIFF
--- a/src/EstablishmentRoll/EstablishmentRoll.js
+++ b/src/EstablishmentRoll/EstablishmentRoll.js
@@ -4,10 +4,9 @@ import { connect } from 'react-redux'
 import axios from 'axios'
 import { setEstablishmentRollData, setApplicationTitle } from '../redux/actions'
 import EstablishmentRollBlock from './EstablishmentRollBlock'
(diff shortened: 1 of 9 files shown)


src/EstablishmentRoll/EstablishmentRoll.js
```diff
@@ -4,10 +4,9 @@ import { connect } from 'react-redux'
 import axios from 'axios'
 import { setEstablishmentRollData, setApplicationTitle } from '../redux/actions'
 import EstablishmentRollBlock from './EstablishmentRollBlock'
-import Spinner from '../Spinner'
-import Error from '../Error'
+import Page from '../Page/Page'
 
-export class EstablishmentRollContainer extends Component {
+export class EstablishmentRoll extends Component {
```
lbennett-moj: I thought we had a general convention for having container in a container name - just interested in knowing why you renamed it.

PR #213, opened 2018-12-04
NN-1439 Add Date of Birth filter

bell-pa:


src/GlobalSearch/DateOfBirth.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+    this.onChange = this.onChange.bind(this)
+    this.errorText = this.errorText.bind(this)
+  }
+
+  componentDidMount() {
+    this.onChange({})
+  }
+
+  onChange(value) {
```
jsrobertson:
```suggestion
  onChange = value => {
```

jsrobertson: Not a deal breaker but using class properties like ^ means you don't have to bind it in the constructor.

package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@govuk-react/constants": "^0.3.1",
     "@govuk-react/header": "^0.3.1",
+    "@govuk-react/date-field": "^0.3.1",
```
jsrobertson: 👍

src/GlobalSearch/DateOfBirth.js
bell-pa: Done

PR #216, opened 2018-12-06
NN-1439: Fix react console warning due to controlled component

petergphillips:


package.json
```diff
@@ -80,6 +80,7 @@
     "html-webpack-plugin": "^3.2.0",
     "husky": "^1.1.2",
     "jest": "^22.4.3",
+    "jsdom": "^13.0.0",
```
petergphillips: don't thinnk this is technically required since we are running jest with `jsdom` option, however linting moans about it

src/GlobalSearch/DateOfBirth.test.js
```diff
@@ -1,37 +1,38 @@
...
 
+const jsdom = new JSDOM('<!doctype html><html lang="en"><body></body></html>')
+const { window } = jsdom
+global.window = window
+global.document = window.document
+
 describe('DateOfBirth', () => {
   it('Should render', () => {
-    shallow(<DateOfBirth />)
+    mount(<DateOfBirth />)
```
petergphillips: enzyme only supports `ref`s for mounted components

src/GlobalSearch/DateOfBirth.test.js
```diff
@@ -1,37 +1,38 @@
...
-      .simulate('change', { target: { value: '1' } })
+    const textinputs = dob.find('input[type="number"]')
+
+    textinputs.at(0).simulate('change', { target: { value: '1' } })
     expect(handleDateOfBirthChange).toHaveBeenCalledWith({ valid: false, blank: false })
-    dob
-      .find('input[type="number"]')
-      .at(1)
-      .simulate('change', { target: { value: '1' } })
+    textinputs.at(0).instance().value = '1'
```
petergphillips: `simulate` doesn't actually change the input value, hence have to amend the value too

PR #244, opened 2018-12-19
Nn-1500 Add clickable 'Flags' to the 'In today' results page

bell-pa:


JamieDraperUK:
Not sure about the way we’re unit testing the alerts markup, feels brittle. Otherwise LFGTM!

PR #247, opened 2019-01-02
Intermediate fix for the issue around requests for bundle.js.map

JamieDraperUK:
…ing the session to expire


JamieDraperUK:
The issue is around a request for bundle.js.map being made and not being authenticated, forcing the current session to expire while dev tools are open.  The fix is to remove it until we can find a better solution.

PR #255, opened 2019-01-16
NN 1400 en route result page

JamieDraperUK:


src/EnRoute/EnRoute.js
```diff
@@ -0,0 +1,94 @@
...
+            </td>
+          </tr>
+        ))}
+      </tbody>
+    </table>
+  </React.Fragment>
+)
+
+EnRoute.propTypes = {
+  rows: PropTypes.arrayOf(
```
jsrobertson: Is this used or likely to be used anywhere else? Could be a contender for the types.js file if so.

JamieDraperUK: All three result pages have slightly different properties. Maybe we could create some base type and reuse some of the fields. Will look at this during the implementation of the next result page.

src/EnRoute/EnRouteContainer.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  render() {
+    const { offenders, loaded } = this.state
+    const { error, agencyId, caseLoadOptions } = this.props
+    const { description } = caseLoadOptions.find(cl => cl.caseLoadId === agencyId)
+    const pageTitle = `En route to ${description}`
+
+    return (
+      <Page title={pageTitle} error={error} loaded={loaded}>
```
jsrobertson: Are the error and loaded props needed?  <Page> should deal with that

PR #267, opened 2019-01-22
NN-1583 add new movement filter parameters

JamieDraperUK:


JamieDraperUK:
Closing because the solution doesn't solve the issues with city locations.

PR #270, opened 2019-01-28
added global-search url, show different results title if no searchtex…

whitfield-mj:
…t, link to licences if licences user

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -46,7 +46,7 @@ const axios = require('axios')
 
 class App extends React.Component {
   async componentWillMount() {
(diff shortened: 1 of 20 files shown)


src/App.js
```diff
@@ -63,15 +63,19 @@ class App extends React.Component {
     )
 
     try {
-      this.loadUserAndCaseload()
```
whitfield-mj: I would suggest this is not done in componentWillMount in general but noticed it is the standard way. Would also probably move the api calls to an async action/saga

PR #276, opened 2019-01-30
fixed typo in licences role type

whitfield-mj:


src/GlobalSearch/GlobalSearchContainer.test.js
```diff
@@ -56,12 +55,10 @@ describe('Global search container', () => {
   })
 
   it('should pass licencesUser true if roles includes LICENCES_RO', async () => {
```
petergphillips: missed one :grin:

PR #285, opened 2019-02-15
NN-1615 appointment form tweaks

JamieDraperUK:


src/BulkAppointments/AppointmentForm.js
```diff
@@ -34,32 +34,32 @@ export const onSubmit = onSucces => values => {
...
+    formErrors.push({ targetName: 'location', text: 'Select location' })
   }
 
   if (!values.date) {
     formErrors.push({ targetName: 'date', text: 'Date is required' })
   }
 
   if (isToday && moment(values.startTime).isBefore(now)) {
-    formErrors.push({ targetName: 'startTime', text: "Start time shouldn't be in the past" })
+    formErrors.push({ targetName: 'startTime', text: 'Start time must be in the future' })
```
petergphillips: this message doesn't correspond to the validation.  Start time can be now can't it?

JamieDraperUK: Good point. It was better before. I'll change it back.

petergphillips: although you can't use the word `shouldn't` as it implies a choice, `must` is better

src/BulkAppointments/AppointmentForm.js
```diff
@@ -34,32 +34,32 @@ export const onSubmit = onSucces => values => {
   }
 
   if (!values.appointmentType) {
-    formErrors.push({ targetName: 'appointmentType', text: 'Appointment type is required' })
+    formErrors.push({ targetName: 'appointmentType', text: 'Select appointment type' })
   }
 
   if (!values.location) {
-    formErrors.push({ targetName: 'location', text: 'Location is required' })
+    formErrors.push({ targetName: 'location', text: 'Select location' })
```
petergphillips: I think `Select a location` would fit more with the GDS style messages, they have things like `Enter a postcode`

src/BulkAppointments/AppointmentForm.js
```diff
@@ -34,32 +34,32 @@ export const onSubmit = onSucces => values => {
...
   }
 
   const DATE_TIME_FORMAT_SPEC = 'YYYY-MM-DDTHH:mm:ss'
 
   if (
     moment(values.endTime, DATE_TIME_FORMAT_SPEC).isBefore(moment(values.startTime, DATE_TIME_FORMAT_SPEC), 'minute')
   ) {
-    formErrors.push({ targetName: 'startTime', text: "Start time should't be after End time" })
-    formErrors.push({ targetName: 'endTime', text: "End time shouldn't be before Start time" })
+    formErrors.push({ targetName: 'startTime', text: 'The start time must be before the end time' })
```
petergphillips: you haven't got `The` at the start of any of the other messages, so not needed here.
It is also odd that you are indicating that both fields are in error, whereas it is likely that just the end time is at fault.  I would have just highlighted the end time.

src/BulkAppointments/AppointmentForm.js
```diff
@@ -34,32 +34,32 @@ export const onSubmit = onSucces => values => {
...
+    formErrors.push({ targetName: 'endTime', text: 'The end time must be after the start time' })
```
petergphillips: Can the end time be the same as the start time?  Validation allows it

src/BulkAppointments/AppointmentForm.js
JamieDraperUK: The wording is from Clare. I'll catch up with her to see what she thinks

src/BulkAppointments/AppointmentForm.js
JamieDraperUK: The following text has been supplied by Clare.
'The start time must be before the end time’
’The end time must be after the start time

petergphillips: You don't need both conditions if you reverse the logic so:
```
const isNotAfter = ! moment(values.endTime, DATE_TIME_FORMAT_SPEC).isAfter(
    moment(values.startTime, DATE_TIME_FORMAT_SPEC),
    'minute'
  )
```
although then it becomes clear that the variable should be called `endNotLaterThanStart` or similar

PR #289, opened 2019-02-20
NN-1688: Call new auth endpoints and cope with no active caseload

petergphillips:


backend/api/elite2Api.js
```diff
@@ -15,17 +15,14 @@ const elite2ApiFactory = client => {
 
   const put = (context, url, data) => client.put(context, url, data).then(processResponse(context))
 
-  const userLocations = context => get(context, 'api/users/me/locations')
-  const currentUser = context => get(context, 'api/users/me')
-  const userRoles = context => get(context, 'api/users/me/roles')
-  const userCaseLoads = context => get(context, 'api/users/me/caseLoads')
+  const userLocations = context => (context.authSource !== 'auth' ? get(context, 'api/users/me/locations') : [])
```
petergphillips: would have preferred this to be only if `authSource === 'nomis'`, however that would require all three apps being released at once.  Can do it this way and then change later on

backend/sessionManagementRoutes.js
```diff
@@ -145,7 +145,7 @@ const configureRoutes = ({ app, healthApi, tokenRefresher, mailTo, homeLink }) =
 
   app.get('/autherror', (req, res) => {
     res.status(401)
-    return res.render('pages/autherror', {
+    return res.render('autherror', {
```
petergphillips: wrong location for autherror

src/GlobalSearch/GlobalSearchResultList.js
```diff
@@ -57,7 +57,7 @@ const GlobalSearchResultList = ({
             <img
               alt={`prisoner ${prisoner.offenderNo}`}
               id={`image-${prisoner.offenderNo}`}
-              className="photo clickable"
+              className="photo"
```
petergphillips: no idea why image was marked as clickable when no enclosing `<a>` present

PR #305, opened 2019-03-11
NN-1705 start time overwritten bug and scrolling issue

JamieDraperUK:


src/BulkAppointments/AddPrisoners/AddPrisonersContainer.js
```diff
@@ -23,6 +23,7 @@ import routePaths from '../../routePaths'
 import OffenderUpload from '../OffenderUpload/OffenderUpload'
 import AddPrisoners from './AddPrisoners'
 import AppointmentDetails from '../../AppointmentDetails/AppointmentDetails'
+import ScrollToTop from '../../Components/ScrollToTop/ScrollToTop'
```
jsrobertson:
```suggestion
import ScrollToTop from '../../Components/ScrollToTop/'
```

src/BulkAppointments/AddPrisoners/AddPrisonersContainer.js
```diff
@@ -97,7 +98,7 @@ class AddPrisonerContainer extends Component {
     return (
       <Page title="Add prisoners" applicationTitle="Whereabouts" alwaysRender>
         {complete && (
-          <div>
+          <ScrollToTop always>
             <H3> {appointmentsCreated} appointments created </H3>
```
jsrobertson: @govuk-react/heading?

JamieDraperUK: At the moment it uses @govuk-react/heading. I just went with what Page.js uses.  I did try and upgrade to use heading as header is decrepcated now but was met with a few issues not resolving 'styled-components'. We need to upgrade to the latest gov components but maybe not for this ticket?

JamieDraperUK: Will replace heading with header as it's deprecated.

PR #311, opened 2019-03-18
NN-859 Recurring appointments off by one fix and GA events

JamieDraperUK:


src/BulkAppointments/RecurringAppointments.js
```diff
@@ -12,7 +14,9 @@ const calculateEndDate = ({ startTime, repeats, numberOfTimes }) => {
   }
 
   const unitOfTime = repeatsToMomentMap[repeats]
-  const times = Number(repeats === 'FORTNIGHTLY' ? Number(numberOfTimes) * 2 : Number(numberOfTimes))
+
+  const InclusiveAdjustment = numberOfTimes - 1
```
petergphillips: lowercase variable should start with lowercase?

JamieDraperUK: Ah good spot

JamieDraperUK: Ah good spot

PR #324, opened 2019-03-21
NN-859 Strip out offenders not in active caseload

JamieDraperUK:


src/BulkAppointments/AddPrisoners/__snapshots__/AddPrisonerContainer.test.js.snap
```diff
@@ -148,6 +148,7 @@ ShallowWrapper {
             />
```
JamieDraperUK: Planning on moving to react-test-renderer but not for this commit

PR #330, opened 2019-03-26
NN-1665 Add results per page dropdown to global search

jsrobertson:


src/Components/ResultsFilter/elements/PerPageDropdown/PerPageDropdown.js
```diff
@@ -0,0 +1,38 @@
...
+import { SelectInput } from '@govuk-react/select'
+
+const DropdownContainer = styled.div`
+  span {
+    font-weight: 700;
+  }
+`
+
+const PerPageDropdown = ({ handleChange, totalResults, perPage }) => {
+  const defaultValue = totalResults <= 20 && perPage !== 10 ? totalResults : perPage
```
JamieDraperUK:
```suggestion
 const defaultValue = Math.min(totalResults, perPage)
```

src/Components/ResultsFilter/index.js
```diff
@@ -0,0 +1,6 @@
+import ResultsFilter from './ResultsFilter'
+import PerPageDropdown from './elements/PerPageDropdown'
+
+ResultsFilter.PerPageDropdown = PerPageDropdown
```
JamieDraperUK: Is the dependency necessary?

jsrobertson: Yeah I think so.  It should only ever be used within a <ResultsFilter /> which contains the pagination info and number of Results.

PR #342, opened 2019-04-17
NN-1835 Fix some intermittent integration tests.

bell-pa:


Nimphal:
This was tested locally by adding delays to the relevant elite2 api calls using wiremock's `withFixedDelay` http://wiremock.org/docs/simulating-faults/

PR #348, opened 2019-04-30
NN-1632 Adding viewport and sizing components.

andrewrlee:
These sizing components can be used to view/hide sections based on screen size.

Uses gov-react constants with the breakpoint being 769px.

Elements smaller than this will be hidden if wrapped in DesktopOnly and only displayed if wrapped with MobileOnly.

These are not currently used but will be in follow ups.

### DIFF
--- a/src/Components/Sizing/index.js
+++ b/src/Components/Sizing/index.js
@@ -1,0 +1,17 @@
+import styled from 'styled-components'
+import { MEDIA_QUERIES } from '@govuk-react/constants'
+
+export const MobileOnly = styled.div`
+  display: block;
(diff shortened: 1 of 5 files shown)


src/Components/Sizing/index.js
```diff
@@ -0,0 +1,17 @@
+import styled from 'styled-components'
```
jsrobertson: When making a new component I've generally a `ComponentName.js` file and then import and export it in an `index.js` file.  Mainly for the fact that it helps locate the right file when you've got multiple files open in your IDE.

andrewrlee: I've done that now

PR #351, opened 2019-05-03
NN-1632 Adding endpoint to allow retrieval of adjudications.

andrewrlee:


.eslintrc
```diff
@@ -18,7 +18,8 @@
     "no-unused-vars": [
       1,
       {
-        "argsIgnorePattern": "res|next|^err|_"
+        "argsIgnorePattern": "res|next|^err|_",
+        "ignoreRestSiblings": true
```
andrewrlee: This is to allow creating a copy of an object [excluding certain properties through using rest syntax](https://github.com/eslint/eslint/issues/4880)

backend/controllers/adjudicationHistoryService.js
```diff
@@ -0,0 +1,42 @@
+const moment = require('moment')
+
+const AdjudciationHistoryServiceFactory = elite2Api => {
+  const getAdjudications = async (context, offenderNumber, params) => {
+    const { requestHeaders, ...withoutPagination } = context
+
+    const [adjudications, findingTypes] = await Promise.all([
+      elite2Api.getAdjudications(context, offenderNumber, params),
+      elite2Api.getAdjudicationFindingTypes(withoutPagination),
+    ])
```
andrewrlee: The context allows headers from the request downstream to pass through to api calls made by the backend.
It also allows paging headers from the response to be passed back to the frontend. 

This works great when calling down to a single endpoint that supports pagination, but is a bit of a gotcha when making more than one call to these endpoints.
  
In this case, I don't want to apply the pagination headers to the `getAdjudicationFindingTypes` request and also don't want the pagination headers from this request to be returned to the frontend. 

As firing off these requests is asynchronous, the version of headers that get written back to the response is non-deterministic, so this works by creating a copy of the context without the pagination, and the modifications to this copy of ignored as they are irrelevant here.

PR #354, opened 2019-05-08
IEP history view (no form filtering yet)

Nimphal:
Part of NN-1631 (no form filtering, breadcrumbs need some TLC).

There are a few steps to getting this data on the page:
1. Get the booking id associated with an offender number (as well as other details)
2. Get the IEP data associated with the booking. 
3. Loop over the data and get the Establishment description through the agency Id in the IEP data and the Staff name through the user id in the IEP data.
4. Calculate next review date (one year from date of last IEP record)

This page isn't linked-to from anywhere so I though it would be wise to get the first part reviewed before I proceed with the form filtering.

There are a couple of es6-linter rules I had to turn off in `backend/controllers/iepHistory.js` in order to get looping over the data and doing extra elite2 api calls to work in the right order. I am suspicious of turning off these rules, I am sure I am missing a trick.

screenshot-localhost-3002-2019 05 08-15-09-37


backend/controllers/iepHistory.js
```diff
@@ -0,0 +1,50 @@
+const moment = require('moment')
+
+const getIepHistoryFactory = elite2Api => {
+  const getIepHistory = async (context, offenderNo) => {
+    const bookingDetails = await elite2Api.getDetails(context, offenderNo)
+    const iepSummary = await elite2Api.getIepSummaryWithDetails(context, bookingDetails.bookingId)
+    const iepHistoryDetails = []
+
+    /* eslint-disable no-restricted-syntax */
```
JamieDraperUK: The current coding style is more functional. See example https://github.com/ministryofjustice/prisonstaffhub/commit/3475ab3590df98e78b5be6197a6d5ff2180d26b8

backend/controllers/iepHistory.js
```diff
@@ -0,0 +1,50 @@
...
+    for (const details of iepSummary.iepDetails) {
+      /* eslint-disable no-await-in-loop */
+      const iepEstablishment = await elite2Api.getAgencyDetails(context, details.agencyId)
```
andrewrlee: I would have thought that an offender is likely to have many ieps at the same agency? - you could probably significantly reduce the number of api calls here by getting a list of the distinct agencyIds and then doing the call once for each agency id?

prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/specs/IepHistorySpecification.groovy
```diff
@@ -0,0 +1,86 @@
...
+        elite2api.stubOffenderDetails('A1234AC', offenderDetails)
+        elite2api.stubAgencyDetails('LEI', agencyDetails)
+        elite2api.stubUserDetails('ITAG_USER', userDetails)
+        elite2api.stubGetIepSummaryWithDetails('-3')
+
+        given: "I am logged in"
+        fixture.loginAs(ITAG_USER)
+
+        when: "I view the IEP history page"
+        go "/offenders/A1234AC/iep-history"
```
andrewrlee: I think to use the `url` variable in the `IepHistory` page you can do:

 `to IepHistory` instead of the `go`

According to [this](http://www.gebish.org/manual/current/#scripting-with-page-objects) it will also do the `at IepHistory` bit too.

backend/controllers/iepHistory.js
Nimphal: You are very right! I have created unique sets of the values and only issue the call once per value.

prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/specs/IepHistorySpecification.groovy
Nimphal: Thank you, have changed it to use that!

backend/controllers/iepHistory.js
Nimphal: That's brilliant, thank you so much! I made a some small changes to make sure it only called the apis once per agency/user, but other than that it worked pretty much out of the box. Thank you so much for taking the time!

src/DateTimeFormatter/index.js
```diff
@@ -0,0 +1,13 @@
+import PropTypes from 'prop-types'
```
andrewrlee: Jamie added a comment to one of my PRs suggesting separating out an index.js and a Component.js file: https://github.com/ministryofjustice/prisonstaffhub/pull/348#discussion_r279771603

backend/controllers/iepHistory.js
```diff
@@ -0,0 +1,56 @@
...
+
+const getIepHistoryFactory = elite2Api => {
+  const getIepHistory = async (context, offenderNo) => {
+    const bookingDetails = await elite2Api.getDetails(context, offenderNo)
+    const iepSummary = await elite2Api.getIepSummaryWithDetails(context, bookingDetails.bookingId)
+
+    // Offenders are likely to have multiple IEPs at the same agency.
+    // By getting a unique list of users and agencies, we reduce the duplicate
+    // calls to the database.
+    const uniqueUserIds = [...new Set(iepSummary.iepDetails.map(details => details.userId))]
```
andrewrlee: think you might be able to drop the first summary detail by doing something like this up here:
`const [summaryDetail, ...iepDetails] =  iepSummary.iepDetails;`

and then refer to iepDetails in all the mapping below... might be a bit more of an explicit way of ignoring the first record, and means you don't need to mutate state with the shift below.

PR #356, opened 2019-05-13
Nn 1632 breadcrumb

andrewrlee:
Provided a `WithOffenderDetails` component that can wrap pages with an offender in context and populate the offender's details in global state. I've then added a new OffenderBreadcrumb which reads the offender details out and renders the name in the crumb correctly.


src/Components/Breadcrumb/Breadcrumb.js
```diff
@@ -10,7 +10,7 @@ export const Breadcrumb = ({ breadcrumbs, homeLink }) => {
   const { breadcrumb: poppedBreadcrumb } = breadcrumbs.length > 0 ? breadcrumbs.pop() : breadcrumbs
 
   return (
-    <BreadcrumbContainer>
+    <BreadcrumbContainer id="breadcrumb">
```
jsrobertson: If this is only used for int testing then, could we use a `data-qa` attribute?

andrewrlee: Now using data-qa instead

PR #357, opened 2019-05-13
NN-1632 Adding date validation to adjudication history filtering

andrewrlee:
Changed `Date Range` Label to `Reported Date`

Added date range validation to prevent filtering on nonsensical date ranges - where `From date` is greater than the `To date`. 

Screenshot 2019-05-13 at 11 20 20

Also Renamed AddPrisonerValidation to match other AddPrisoners file names


src/Adjudications/AdjudicationHistory/AdjudicationHistoryFormValidation.js
```diff
@@ -0,0 +1,21 @@
+import { FORM_ERROR } from 'final-form'
+
+const validateThenSubmit = ({ fromDate, toDate, submit }) => values => {
+  if (toDate && fromDate && fromDate.isAfter(toDate)) {
+    return {
+      [FORM_ERROR]: [
+        {
+          targetName: 'fromDate',
+          text: 'The From date must be the same or before the To date',
```
petergphillips: missing `as` i.e. should be:
```The From date must be the same as or before the To date```
Same with `toDate`

File: src/Adjudications/AdjudicationHistory/AdjudicationHistoryFormValidation.js [hunk 2]
```diff
@@ -0,0 +1,21 @@
+import { FORM_ERROR } from 'final-form'
+
+const validateThenSubmit = ({ fromDate, toDate, submit }) => values => {
+  if (toDate && fromDate && fromDate.isAfter(toDate)) {
```
petergphillips: will you be adding the other validation at a later date e.g. dates are required, `toDate` must be in the past etc...

src/Adjudications/AdjudicationHistory/AdjudicationHistoryFormValidation.js
andrewrlee: fixed this

src/Adjudications/AdjudicationHistory/AdjudicationHistoryFormValidation.js
andrewrlee: as discussed - other validation isn't required as its click based so you can't put in invalid fields

PR #358, opened 2019-05-14
NN-1632 Adding offender page titles

andrewrlee:
Converted `WithOffenderDetails` in to an `OffenderPage`. This allows creating a title from offender details and removes the title when an error occurs.

Also fixed an issue with OffenderBreadcrumb so it display nicely when an unexpected error occurs.

Screenshot 2019-05-14 at 11 54 52


andrewrlee:
This is going to need to wait until Nev's refactor as it breaks the IEP breadcrumb.

backend/csv-parser.js
```diff
@@ -10,8 +10,8 @@ const validationMessages = {
 const csvParserService = ({ fs, isBinaryFileSync }) => {
   const parseCsvData = async data => {
     const output = []
-
-    const parser = parse(data, { trim: true, skip_empty_lines: true }).on('readable', function() {
+    // eslint-disable-next-line prefer-arrow-callback
+    const parser = parse(data, { trim: true, skip_empty_lines: true }).on('readable', function onRead() {
```
andrewrlee: had some help from Jamie to remove this warning from the linter

prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/pages/IepHistory.groovy
```diff
@@ -6,7 +6,7 @@ class IepHistory  extends Page {
     static url = '/offenders/A1234AC/iep-level'
 
     static at = {
-        pageTitle == 'IEP History for NORMAN BATES'
+        pageTitle == 'IEP History for Norman Bates'
```
andrewrlee: @Nimphal can you confirm you are ok with this? 

This matches how we do things for scheduled events: 
Screenshot 2019-05-15 at 09 31 06

Nimphal: Yeap, happy.

andrewrlee: I'll change the `history` to be lower case

src/Adjudications/AdjudicationHistory/AdjudicationHistoryContainer.test.js
```diff
@@ -37,6 +37,6 @@ describe('Adjudication History container', () => {
       />
     )
 
-    expect(wrapper.dive()).toMatchSnapshot()
+    expect(wrapper.dive().debug()).toMatchSnapshot()
```
jsrobertson: Not sure about `.debug()`.  Generally would only try to use this when debugging.  I still think we should probably be using react-test-renderer for snapshot stuff.

src/Components/OffenderBreadcrumb/OffenderBreadcrumb.test.js
```diff
@@ -1,11 +1,19 @@
 import React from 'react'
-import { shallow } from 'enzyme'
+import ShallowRenderer from 'react-test-renderer/shallow'
```
jsrobertson: Curious to know why you've switched enzyme out.

andrewrlee: Jamie suggested that we should be moving away from `enzyme` and moving to `react-test-renderer` where possible.

I've been a bit inconsistent though within this PR, mostly due to limitations with the shallow rendering in `react-test-renderer`.
 * It doesn't call `onComponentMount`  
 * It only renders the top most element and you can't dive, like in enzyme.
 
I'm not totally convinced that there is a one size fits all approach to this.

src/Adjudications/AdjudicationHistory/AdjudicationHistoryContainer.test.js
andrewrlee: Found one of the enzyme developers providing it as a possible solution here: https://github.com/airbnb/enzyme/issues/1533#issuecomment-367392996

A bit reluctant to do a whole render as it would break encapsulation of the wrapped components - I'd have to put in loads of info into this test about how to set up the descendent's item state correctly (It wraps `Page` which requires quite a lot of state). 

The snapshot would also then contain all the information about how they are rendered - so any changes to them would break this test. This test is attempting to just test the structure of this component - the other components are tested in their own unit tests.

andrewrlee: It seems like jest-24 requires [custom](https://github.com/facebook/jest/issues/7802#issuecomment-460604154) [serializer](https://github.com/kentcdodds/jest-glamor-react) registering. That would allow moving away from debug

PR #359, opened 2019-05-14
Nn 1631 iep history

Nimphal:
Refactors IEP history page to use redux, introduces filtering and abstracts some common parts of IEP history and Adjudication history.


src/IepHistory/IepHistory.test.js
```diff
@@ -1,21 +1,53 @@
...
+  })
+
+  it('should render the iep history table correctly', () => {
+    const wrapper = shallow(<IepHistory store={store} />)
+
+    expect(wrapper).toMatchSnapshot()
+  })
+
+  it('should render the current IEP level correctly', () => {
+    const wrapper = shallow(<CurrentIepLevel store={store} />)
```
JamieDraperUK: We've been trying to move towards https://reactjs.org/docs/test-renderer.html 

See example in AddPrisoners.test.js

backend/controllers/iepHistory.js
```diff
@@ -1,7 +1,7 @@
 const moment = require('moment')
 
 const getIepHistoryFactory = elite2Api => {
-  const getIepHistory = async (context, offenderNo) => {
+  const getIepHistory = async (context, offenderNo, params) => {
```
JamieDraperUK: It would be good to do some validation here. The reason I say this is we've noticed that in production it that the client side validation doesn't seem to kick in.  

This should help minimise the logs across all services.

backend/controllers/iepHistory.js
```diff
@@ -10,6 +10,7 @@ const getIepHistoryFactory = elite2Api => {
     // calls to the database.
     const uniqueUserIds = [...new Set(iepSummary.iepDetails.map(details => details.userId))]
     const uniqueAgencyIds = [...new Set(iepSummary.iepDetails.map(details => details.agencyId))]
+    const levels = [...new Set(iepSummary.iepDetails.map(details => details.iepLevel))]
```
andrewrlee: Could possibly extract out a separate variable that explicitly ignores the first detail?:
`const [summaryDetail, ...iepDetails] = iepSummary.iepDetails` 

It might be a bit more explicit, and then you could refer to `iepDetails` throughout and avoid the mutation of state with the `shift`.

backend/controllers/iepHistory.js
```diff
@@ -38,13 +39,35 @@ const getIepHistoryFactory = elite2Api => {
 
     iepHistoryDetails.shift()
 
+    let filteredResults = iepHistoryDetails
```
JamieDraperUK: Definitely, personal preference but might be cleaner to have a filtering logic in its own function, this might remove the need for the let keyword.

backend/controllers/iepHistory.js
Nimphal: Yeap, have done this, thank you!

backend/controllers/iepHistory.js
Nimphal: Talked this over in person, form data is not used in the elite2 api calls, so there is no need for validation here.

backend/controllers/iepHistory.js
Nimphal: My preference too, have done so :)

backend/controllers/iepHistory.js
```diff
@@ -10,6 +33,7 @@ const getIepHistoryFactory = elite2Api => {
     // calls to the database.
     const uniqueUserIds = [...new Set(iepSummary.iepDetails.map(details => details.userId))]
     const uniqueAgencyIds = [...new Set(iepSummary.iepDetails.map(details => details.agencyId))]
```
andrewrlee: do you need any sorting on establishments or levels? or is the order in which they are presented in the select specified in the web layer?  

(There might be a very slight quirk around having an extra establishment or level in the dropdown which doesn't match any of the details if the summary detail has a level or establishment that isn't shared by any of the other details.. it's very minor, could possibly be fixed by moving the `const [summaryDetail, ...iepDetails]`to just after the data load and using it in each of these maps.)

PR #360, opened 2019-05-15
NN-1560 Added paid and other to Activity Lists and use the new Whereabouts API

jsrobertson:
Still WIP but it's behind the `UPDATE_ATTENDANCE_ENABLED` feature.

Todo:
- Move the absent reasons into redux state rather than at component level
- Component tests for PayOtherForm and PayOptions


src/ResultsActivity/elements/PayOptions.js
```diff
@@ -0,0 +1,98 @@
...
+          <VisuallyHidden>Pay</VisuallyHidden>
+        </Radio>
+      </Option>
+
+      <Option>
+        <Radio
+          name={offenderNo}
+          onChange={otherHandler}
+          value="other"
+          data-first-name={firstName}
```
jsrobertson: Decided on a better way to do this.  Will update the PR tomorrow.

backend/controllers/activityList.js
```diff
@@ -51,6 +71,18 @@ const getActivityListFactory = elite2Api => {
 
     const offenderNumbersWithDuplicates = eventsAtLocation.map(event => event.offenderNo)
     const offenderNumbers = [...new Set(offenderNumbersWithDuplicates)]
+    const offenderBookingIds = await Promise.all(
```
JamieDraperUK: This is only needed for the attendance calls, maybe this needs to be placed behind a feature flag too?

PR #364, opened 2019-05-17
Nn 1921 update filters and display

Nimphal:
Minor interface tweaks to IEP details page:

1. Change `IEP history/level` to `IEP details` in page heading, breadcrumbs and urls.
2. Fiddle with padding on form to reduce empty grey space.
3. User lighter grey colour for form background to match other filter forms on Quick Look such as Case Notes.
4. Increase size of `IEP history` heading
5. Enable `All` option in filters

screenshot-localhost-3002-2019 05 16-15-03-29

### DIFF
--- a/src/Components/FilterForm/FilterForm.styles.js
+++ b/src/Components/FilterForm/FilterForm.styles.js
@@ -14,9 +14,10 @@ export const FullWidthSelect = styled(Select)`
 `
 
 export const SearchArea = styled.div`
-  background: ${GREY_3};
+  background: #f2f2f2;
   padding: ${SPACING.SCALE_5};
-  margin-bottom: ${SPACING.SCALE_3};
+  padding-bottom: ${SPACING.SCALE_3};
+  margin-bottom: ${SPACING.SCALE_5};
 `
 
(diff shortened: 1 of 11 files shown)


src/Components/FilterForm/FilterForm.styles.js
```diff
@@ -14,9 +14,10 @@ export const FullWidthSelect = styled(Select)`
 `
 
 export const SearchArea = styled.div`
-  background: ${GREY_3};
+  background: #f2f2f2;
```
jsrobertson: We should be using colour values from `govuk-colours` where possible.

Nimphal: Yes, this seems to be a conflict between what the ticket specifies and what's available from the gov-uk colours. This shade of grey isn't available in the greys provided by gov-uk, but is used in quick look and has been requested for this story :/

jsrobertson: OK, will arrange a conversation with Eddie to get the consistency right between the apps.  If we are set on having our own colours for components like this in future, we could add our own colour palette to new-nomis-shared (not suggesting we do this right now though).

PR #370, opened 2019-05-21
NN-1927: Update case switching for IEP details and Adjudications

Nimphal:
Currently both of the above pages redirect to wherearebouts home,
which is not the where the user got to them. This updates the switchCaseload
logic to detect pages under the 'offenders' path and to redirect those
to the notm interface instead.

### DIFF
--- a/prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/specs/AgencySelectionSpecification.groovy
+++ b/prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/specs/AgencySelectionSpecification.groovy
@@ -5,10 +5,13 @@ import com.github.tomakehurst.wiremock.junit.WireMockRule
 import org.junit.Rule
 import uk.gov.justice.digital.hmpps.prisonstaffhub.mockapis.Elite2Api
 import uk.gov.justice.digital.hmpps.prisonstaffhub.mockapis.OauthApi
+import uk.gov.justice.digital.hmpps.prisonstaffhub.mockapis.mockResponses.AdjudicationResponses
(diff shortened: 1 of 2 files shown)


prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/specs/AgencySelectionSpecification.groovy
```diff
@@ -45,8 +48,89 @@ class AgencySelectionSpecification extends BrowserReportingSpec {
         elite2api.stubUpdateActiveCaseload()
         notmServer.stubFor(
                 get(WireMock.urlPathMatching('/.*'))
-                .willReturn(
-                    aResponse().withStatus(200)))
+                        .willReturn(
+                                aResponse().withStatus(200)))
+
+        $('#info-wrapper').click()
```
JamieDraperUK: Low-level details like element names and ids should really reside in the page objects.

src/App.js
```diff
@@ -189,9 +189,11 @@ class App extends React.Component {
 
   switchCaseLoad = async (newCaseload, location) => {
```
JamieDraperUK: Definitely need a more general solution to workflows between our apps. One for later.

prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/specs/AgencySelectionSpecification.groovy
Nimphal: Yeah, I think these aren't part of a page because they are the header. I'll have a think about how to fix those and follow-up.

PR #373, opened 2019-05-22
NN-1939 UI tweaks for IEP history

jsrobertson:
Also fixed incorrect PropType warning for `days`

### DIFF
--- a/src/IepDetails/IepHistory.js
+++ b/src/IepDetails/IepHistory.js
@@ -13,8 +13,8 @@ const IepHistory = ({ results }) => (
         <Table.Row>
           <Table.CellHeader setWidth="20%">Date and time</Table.CellHeader>
           <Table.CellHeader setWidth="15%">IEP description</Table.CellHeader>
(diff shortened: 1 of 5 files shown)


src/IepDetails/CurrentIepLevel.js
```diff
@@ -26,13 +26,13 @@ const CurrentIepLevel = ({ level, days, nextReviewDate }) => (
 
 CurrentIepLevel.propTypes = {
   level: PropTypes.string,
-  days: PropTypes.string,
+  days: PropTypes.number,
   nextReviewDate: PropTypes.string,
 }
```
Nimphal: Oops, approved this before eagle-eyed @andrewrlee spotted that actually, the value here is meant to be a string - this is the '1 year, 2 days' value.

jsrobertson: Andrew has just told me he's fixed the type of value that gets passed, locally, so I've reverted that part of the change.

PR #376, opened 2019-05-23
NN-1944: Add keep alive to requests

petergphillips:


backend/api/oauthEnabledClient.js
```diff
@@ -24,9 +26,18 @@ const errorLogger = error => {
...
  */
 const factory = ({ baseUrl, timeout }) => {
+  const agentOptions = {
+    maxSockets: 100,
+    maxFreeSockets: 10,
+    freeSocketTimeout: 30000,
+  }
+  const keepaliveAgent = baseUrl.startsWith('https')
+    ? { httpsAgent: new HttpsAgent(agentOptions) }
+    : { httpAgent: new Agent(agentOptions) }
```
andrewrlee: does the key need to be `agent` on the left hand side? 
like: https://github.com/node-modules/agentkeepalive

PR #384, opened 2019-05-29
NN-1677

andrewrlee:
Fixing defect around displaying hearings correctly,

Previously:   
<kbd>
Screenshot 2019-05-29 at 13 59 14
</kbd>

Now:
<kbd>
Screenshot 2019-05-29 at 14 10 21
</kbd>

Also upgrading to the latest version of gradle


andrewrlee:
Should probably change case too!

PR #386, opened 2019-05-31
NN-1908 Display pay details

jsrobertson:
To do:
- UI still needs refining with Eddie


src/ResultsActivity/elements/PayDetails/PayDetails.js
```diff
@@ -0,0 +1,42 @@
+import React, { Fragment } from 'react'
+import PropTypes from 'prop-types'
+import { H2 } from '@govuk-react/heading'
+import Paragraph from '@govuk-react/paragraph'
+import ButtonCancel from '../ButtonCancel'
+import { pascalToString } from '../../../utils'
+
+const PayDetails = ({ paid, absentReason, comments, cancelHandler }) => (
+  <Fragment>
```
andrewrlee: is there a disadvantage to using fragment shorthand style for things like this? (`<></>`)

jsrobertson: I/we never used to use them as our version of babel at the time didn't transform them, however as we're on 7+ now, it should be ok. I'm neitheir here nor there on `<React.Fragment>`,`<Fragment>` or `<>` so perhaps one we should all agree on and add it to the upcoming FE guidelines (I'll make a start later today on those) and we can add to them as a team?

PR #394, opened 2019-06-04
NN-1952 Update new-nomis-shared to use @govuk-react/footer

jsrobertson:


jsrobertson:
Updated this as duplicate `h1`s were breaking integration tests.  Updated the groovy tests but also changed the `h1`s to `h2`s [here](https://github.com/govuk-react/govuk-react/pull/671)

PR #398, opened 2019-06-06
Nn 1806 change iep level

Nimphal:
For https://dsdmoj.atlassian.net/browse/NN-1806

This add the ability for users with the MAINTAIN_IEP role to update the IEP for an offender. The most exotic part of this is the level difference calculation which is used to determine which icon to display and in what order to show the levels. The biggest issue there is that an agency could have created its own levels which do not match our expectations. In that case those levels will appear at the end of the list and will not have icons. 

If a user without the role somehow navigates to this page (they should not see the button), they will be redirected to the IEP details page. 

This is behind a feature flag, which has been set to true in dev so the functionality should be visible without any configuration.

screenshot-localhost-3002-2019 06 06-16-55-27


backend/controllers/iepDetails.js
```diff
@@ -74,8 +74,54 @@ const getIepDetailsFactory = elite2Api => {
...
+          diff = 1000
+        }
+        return {
+          title: level.iepDescription,
+          value: level.iepLevel,
+          image: iconForDifference[diff.toString()] ? iconForDifference[diff.toString()] : '',
+          diff,
+        }
+      })
+      .sort((a, b) => (Math.abs(a.diff) > Math.abs(b.diff) ? 1 : -1))
```
andrewrlee: Might not be reading this right, but would have thought the `abs` would mean that if your current level is `basic` then the order of  `entry` or `standard` is  non deterministic as they would both have the same diff? 
(or maybe it is deterministic based on iteration order? - which isn't very explicit)

src/App.scss
```diff
@@ -177,6 +181,10 @@ table tr:nth-child(even){
   margin-left: 17px;
 }
 
+.margin-left-medium {
+  margin-left: 10px;
+}
```
andrewrlee: should we use styled components for the elements using these?

backend/controllers/iepDetails.js
Nimphal: You are right, they will have the same diff, and that is intentional as the diff represents the delta for ordering purposes.

Generally we want the ordering to be 1 step (up or down), 2 steps (up or down), 3 steps (only up possible as Entry cannot be assigned from this page). When it's two 1 step changes we want the delta to be the same for ordering purposes. (I am not sure I am explaining this right...)

andrewrlee: oh.. maybe that's not a good example as `entry` is filtered out - but maybe if the current level is `standard` and the possible levels are `basic` and `enhanced`

Nimphal: Yeah, the above applies to Standard change to Basic and Enhanced - the delta between them is one, and we're ordering by deltas - it doesn't really matter which order those two appear in as their deltas are identical (they appear alphabetically currently). 

The absolute is there to account for two negative steps. For example, current level is Enhanced, the two options are Standard (diff: -1) and Basic (diff: -2). Standard is one delta down, Basic is two deltas down. However, if we compared them by difference `-1 > -2`, so Basic would appear first even though it has the higher delta.

This broke my brain for quite a long time...

src/App.scss
Nimphal: We totally should, have done so.

backend/controllers/iepDetails.js
```diff
@@ -74,8 +74,54 @@ const getIepDetailsFactory = elite2Api => {
...
+        if (levelToIntMap[level.iepDescription] && levelToIntMap[currentIepLevel]) {
+          diff = levelToIntMap[level.iepDescription] - levelToIntMap[currentIepLevel]
+        } else {
+          // This is a custom level for which we do not have an icon. Always show it last.
+          diff = 1000
+        }
+        return {
+          title: level.iepDescription,
+          value: level.iepLevel,
+          image: iconForDifference[diff.toString()] ? iconForDifference[diff.toString()] : '',
```
andrewrlee: meh... think you could do:   `iconForDifference[diff.toString()] || '',`

backend/controllers/iepDetails.js
andrewrlee: cool, that makes sense to me.

Wonder if it would be a bit more explicit to order by the magnitude from the `levelToIntMap` instead.

Nimphal: That would work for almost all cases, but if the current level is Enhanced (mag. 4), the two options are Basic (2) and Standard (3). Ordering by ascending will show Basic first (delta = 2) and then Standard (delta 1) because Standard's magnitude is bigger.

Ordering in a descending order, when the level is Basic, the two options are Standard (mag. 2) and Enhanced (mag. 3). Ordering by magnitude descending descending would show Enhanced first (delta = 2) and Standard second (delta = 1).

Perhaps I could make the `Make.abs(a.diff)` more explicit by assigning it a variable and calling it delta? As what this is comparing on is not the same thing as the difference - the difference is used for assigning the right icon and negative numbers make sense there whereas we're looking for the delta to be always positive for ordering purposes.

backend/controllers/iepDetails.js
Nimphal: Yeap, have done so.

backend/controllers/iepDetails.js
Nimphal: Have a delta and put a comment that hopefully makes ordering a bit clearer.

backend/controllers/iepDetails.js
```diff
@@ -74,8 +74,55 @@ const getIepDetailsFactory = elite2Api => {
...
+  }
+
+  const getPossibleLevels = async (context, currentIepLevel, agencyId) => {
+    const levels = await elite2Api.getAgencyIepLevels(context, agencyId)
+
+    return levels
+      .filter(level => level.iepDescription !== currentIepLevel)
+      .filter(level => level.iepDescription !== 'Entry')
+      .map(level => {
+        let diff
```
JamieDraperUK: Might look neater to place the diffCalculator in its own method.

backend/controllers/iepDetails.js
```diff
@@ -74,8 +74,55 @@ const getIepDetailsFactory = elite2Api => {
...
+        } else {
+          // This is a custom level for which we do not have an icon. Always show it last.
+          diff = 1000
+        }
+        return {
+          title: level.iepDescription,
+          value: level.iepLevel,
+          image: iconForDifference[diff.toString()] || '',
+          delta: Math.abs(diff), // Used for ordering so the biggest change is shown last regardless of up or down.
+          diff,
```
JamieDraperUK: Would it make sense to just rename this to delta?

Nimphal: The diff is used to determine which icon to use and the delta is used for ordering, they serve different purposes.

backend/controllers/iepDetails.js
```diff
@@ -74,8 +74,55 @@ const getIepDetailsFactory = elite2Api => {
...
+        }
+      })
+      .sort((a, b) => (Math.abs(a.delta) > Math.abs(b.delta) ? 1 : -1))
```
JamieDraperUK: Is Math.abs required here? We already apply it on line 110.

Nimphal: Good catch! Have cleaned up.

backend/controllers/iepDetails.js
Nimphal: Definitely looks neater, have abstracted.

backend/controllers/iepDetails.js
Nimphal: Have removed delta, we just return levelDifference as part of this as it's used for sorting.

PR #399, opened 2019-06-07
NN-1761 Add total results number to activity and houseblock lists

jsrobertson:
Raised [NN-1986](https://dsdmoj.atlassian.net/browse/NN-1986) while doing this as both lists could be combined into one as they're effectively the same thing and we're having to duplicate work.


src/ResultsActivity/ResultsActivity.js
```diff
@@ -293,7 +309,7 @@ class ResultsActivity extends Component {
           {!offenders || offenders.length === 0 ? (
             <div className="font-small padding-top-large padding-bottom padding-left">No prisoners found</div>
```
JamieDraperUK: Shame about the mixed use of scss and styled here :(

PR #401, opened 2019-06-10
NN-1900 Allow attendance entries to be updated

jsrobertson:


backend/controllers/attendance.js
```diff
@@ -2,42 +2,23 @@ const moment = require('moment')
...
 
-const attendanceFactory = (elite2Api, whereaboutsApi) => {
-  const postAttendance = async (context, attendance) => {
-    if (!attendance || !attendance.offenderNo) {
-      throw new Error('Offender number missing')
+const attendanceFactory = whereaboutsApi => {
+  const updateAttendance = async (context, attendance) => {
+    if (!attendance || !attendance.bookingId) {
+      throw new Error('Booking ID is missing')
     }
```
Nimphal: Is there a reason we're not showing an error message if an attendance is missing?

PR #407, opened 2019-06-12
NN-1998 Add payment progress indicator

jsrobertson:
As mentioned at the top of `PayOptions.test.js`, some of the tests can be improved with the release of React 16.9.0 with the support of the async `act()` function.  A combination of using the useState hooks and async/await within PayOptions means there will be a couple of harmless console errors  within the tests until 16.9.0, and a couple of the tests (around showing the spinner and the pay radio) are not quite the way I'd like them to be.


src/ResultsActivity/elements/PayOptions/PayOptions.js
```diff
@@ -52,12 +56,14 @@ function PayOptions({ offenderDetails, updateOffenderAttendance, date, openModal
...
-        {offenderNo &&
+        {!isPaying &&
+          offenderNo &&
           eventId && (
             <Radio onChange={payOffender} name={offenderNo} value="pay" checked={selectedOption === 'pay'}>
               <VisuallyHidden>Pay</VisuallyHidden>
             </Radio>
           )}
+        {isPaying && <Spinner title="Paying" height="40px" width="40px" />}
       </Option>
```
Nimphal: Should we use govuk spacing here? I think it may be 7 or something...

jsrobertson: Yeah you're right @Nimphal .  It's supposed to match the size of the radio buttons, which use spacing point 7.  Will update, cheers.

src/ResultsActivity/elements/PayOptions/PayOptions.test.js
```diff
@@ -26,26 +29,39 @@ describe('<PayOptions />', () => {
...
   it('should call updateOffenderAttendance with the correct args when selecting pay', () => {
-    payRadio.props.onChange()
+    act(() => payRadio.props.onChange())
     expect(props.updateOffenderAttendance).toHaveBeenCalledWith(
       { attended: true, eventId: 123, eventLocationId: 1, offenderNo: 'ABC123', paid: true },
       1
     )
   })
 
+  it('should load the paying spinner when selecting pay', () => {
```
jsrobertson: With 16.9.0 we'll be able to change this to something like:
```
it('should load the paying spinner when selecting pay', async () => {
    act(async () => {
      await payRadio.props.onChange()
    })
    expect(testInstance.findByType(Spinner))
    await Promise.resolve
    expect(testInstance.findByProps({ 'data-qa': 'pay-option' }).findByType(Radio))
  })
```

PR #419, opened 2019-06-28
NN-2014 Show absent reason on printed activity list

jsrobertson:


src/ResultsActivity/ResultsActivity.js
```diff
@@ -241,6 +241,8 @@ class ResultsActivity extends Component {
           attendanceInfo,
         }
 
+        const { absentReason } = attendanceInfo || {}
```
JamieDraperUK: Should the default object live in defaultProps?

PR #436, opened 2019-07-22
NN-1909 Added IEP created slip

jsrobertson:
- Moved modal into `App.js` and created new action to set modal status and content
- Added IEP Created modal with print button
- Added IEP slip which displays when clicking print
- One `<ButtonCancel />` and `<ButtonContainer />`  component which replaces all the other individual instances


src/IEPSlip/IEPSlip.js
```diff
@@ -0,0 +1,113 @@
...
+          Date: <Value inline>{raisedDate.format('D MMMM YYYY')}</Value>
+        </Column>
+      </Row>
+    </IEPSlipDetails>
+  )
+}
+
+IEPSlip.propTypes = {
+  user: PropTypes.shape({}).isRequired,
+  raisedDate: PropTypes.shape({}), // moment object
```
JamieDraperUK: I think we have a proper moment object for this purpose or at least a library.

jsrobertson: Meant to add those, cheers. We have one for user too.

src/ResultsActivity/elements/PayOtherForm/PayOtherForm.js
```diff
@@ -10,47 +9,56 @@ import TextArea from '@govuk-react/text-area'
...
+export function PayOtherForm({ user, offender, updateOffenderAttendance, absentReasons, showModal, activityName }) {
+  const { offenderNo, bookingId, eventId, eventLocationId, attendanceInfo } = offender
+  const { id, absentReason, comments } = attendanceInfo || {}
 
-  if (!values.pay) {
-    formErrors.push({ targetName: 'pay', text: 'Select a pay option' })
-  }
+  const validateThenSubmit = submitHandler => async values => {
+    const formErrors = []
+    const shouldCreateIEP = values.absentReason === 'Refused' || values.absentReason === 'UnacceptableAbsence'
```
JamieDraperUK: We should hard code this check. The absent reasons data contains a list of reasons that trigger IEPS

PR #447, opened 2019-07-26
NN-2074 add client side validation on comments

JamieDraperUK:
The limit of 240 applies to both event outcomes and IEP warnings.

### DIFF
--- a/src/ResultsActivity/elements/PayOtherForm/PayOtherForm.test.js
+++ b/src/ResultsActivity/elements/PayOtherForm/PayOtherForm.test.js
@@ -133,7 +133,7 @@ describe('<PayOtherForm />', () => {
         expect(errors.at(0).text()).toEqual('Enter case note')
       })
 
-      it('should show correct maximum length validation message for the event out come text', async () => {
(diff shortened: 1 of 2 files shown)


src/ResultsActivity/elements/PayOtherForm/PayOtherForm.test.js
```diff
@@ -3,6 +3,11 @@ import { mount } from 'enzyme'
 import { PayOtherForm } from './PayOtherForm'
 import IEPCreated from '../../../IEPCreated'
 
+const buildLargeText = length =>
+  Array.from([...Array(length).keys()])
+    .map(_ => 'A')
+    .join('')
```
andrewrlee: totally meh, could you use [repeat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat)?  `length => 'A'.repeat(length)`

JamieDraperUK: Nice, didn't know you could do that!

PR #448, opened 2019-07-29
NN-2066 pay count

JamieDraperUK:


src/redux/reducers/index.js
```diff
@@ -23,6 +23,11 @@ export function defaultPeriod(time) {
   return 'ED'
 }
 
+const countPaid = data =>
+  data
+    .map(event => (event && event.attendanceInfo && event.attendanceInfo.paid ? 1 : 0))
```
jsrobertson: Not sure it needs `event &&` does it?

PR #450, opened 2019-07-30
Nn 2059 houseblock list attendance values

Nimphal:
For https://dsdmoj.atlassian.net/browse/NN-2059

This adds a bit more complexity in what we show on the houseblock lists. Unlike the activity lists, the absence reason here is not a link so is not editable. Also, if we have no attendance data for a locked record, we should 'Not Recorded'.

### DIFF
--- a/src/ResultsActivity/elements/PayOptions/PayOptions.js
+++ b/src/ResultsActivity/elements/PayOptions/PayOptions.js
@@ -49,12 +49,13 @@ function PayOptions({ offenderDetails, updateOffenderAttendance, date, noPay, sh
     )
 
   const allowUpdate = selectedOption === 'other' && absentReason
+  const notRecorded = !paid && !other && !isWithinLastWeek(date)
   const showRadioButton = !locked && offenderNo && eventId && isWithinLastWeek(date)
 
   const payMessage = () => {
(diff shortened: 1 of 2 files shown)


src/ResultsActivity/elements/PayOptions/PayOptions.js
```diff
@@ -87,10 +88,16 @@ function PayOptions({ offenderDetails, updateOffenderAttendance, date, noPay, sh
...
+          <Fragment>
+            {noPay && <OtherMessage data-qa="other-message">{absentReason.name}</OtherMessage>}
+            {!noPay && (
+              <UpdateLink role="link" onClick={renderForm}>
+                <OtherMessage data-qa="other-message">{absentReason.name}</OtherMessage>
+              </UpdateLink>
+            )}
+          </Fragment>
         )}
+        {!allowUpdate && notRecorded && noPay && <OtherMessage data-qa="other-message">Not Recorded</OtherMessage>}
```
JamieDraperUK: Feel free to ignore this, but could we reduce this to a shouldShowNotRecorded?

Nimphal: I think I might leave it to remind us how wrong this is :D

PR #451, opened 2019-07-30
NN-2066 hide pay total in print

JamieDraperUK:


src/ResultsActivity/ResultsActivity.js
```diff
@@ -37,6 +37,12 @@ const StackedTotals = styled.div`
   flex-direction: column;
 `
 
+const HideForPrint = styled.span`
```
JamieDraperUK: Duplicated, for now, need to clean up the duplication in both the house block and activity lists, either by combining them or moving out common functions and styles out into a common file.

PR #454, opened 2019-07-31
NN-2092 Remove agency prefix from cell location for IEP slip

jsrobertson:


src/IEPCreated/IEPCreated.js
```diff
@@ -24,7 +24,7 @@ const IEPCreated = ({ showModal, offender, iepValues, activityName, user }) => {
       offenderNo: offender.offenderNo,
       offenderName: `${properCaseName(offender.firstName)} ${properCaseName(offender.lastName)}`,
       caseNote: `${pascalToString(iepValues.absentReason)}: ${iepValues.comments}`,
-      cellLocation: offender.cellLocation,
+      cellLocation: stripAgencyPrefix(offender.cellLocation, user.activeCaseLoadId),
```
petergphillips: not sure this looks correct - why use the `user` active case load rather than the `offender.agencyId`?

PR #455, opened 2019-08-01
Nn 2049 add pay all button

Robc24:
New Attend all functionality.  Button to attend all OR all remaining prisoners.


src/ResultsActivity/ResultsActivity.js
```diff
@@ -117,6 +139,66 @@ class ResultsActivity extends Component {
       </div>
     )
 
+    const unpaidOffenders = new Set()
+
+    const attendAllNonAssigned = async () => {
+      try {
+        this.setState({ payingAll: true })
+
```
Nimphal: This would be better as a reducer so it can be tested separately.

src/ResultsActivity/ResultsActivity.js
```diff
@@ -117,6 +139,66 @@ class ResultsActivity extends Component {
...
+        const paidOffenders = response.data.map(offender => {
+          const { paid, attended, bookingId } = offender
+          const offenderAttendanceData = { paid, attended, pay: attended && paid }
+
+          setActivityOffenderAttendance(findOffenderIndexByBookingId(bookingId), offenderAttendanceData)
+
+          return offender
+        })
+
+        this.setState({ payingAll: false })
```
Nimphal: Same as above.

src/ResultsActivity/ResultsActivity.js
```diff
@@ -117,6 +139,66 @@ class ResultsActivity extends Component {
...
+      if (activityData.length === paidList) showControls = false
+
+      const attendanceInfo = activities.filter(activity => activity.attendanceInfo)
+      if (attendanceInfo.length === activities.length) showControls = false
+
+      const lockedCases = attendanceInfo.filter(activity => activity.attendanceInfo.locked === true)
+      if (lockedCases.length === activities.length) showControls = false
+
+      return showControls
+    }
```
Nimphal: Could this be simplified to something like below :
```
return (activityData.length !== paidList) || 
          (attendanceInfo.length !== activities.length) || 
          (lockedCases.length !== activities.length)
```

src/ResultsActivity/ResultsActivity.js
jsrobertson: Not sure a reducer is necessary for setting one piece of local state.

Nimphal: It's more for consistency and testability really, definitely more of a code style point here.

Robc24: I think a reducer for this would be overkill, Nev.  It's just local state for this component rather than global state shared across components.

Nimphal: Happy with that :)

Nimphal:
We might want to consider something in global state that we can use all over the shop when we want to do things like 'Working...' or 'Making magic happen...' (cheesy, I know). Not as part of this PR, but if we something like this again, we should think about abstracting it so it logs properly in console and we can track pages in flight state.

src/ResultsActivity/ResultsActivity.js
```diff
@@ -380,9 +481,14 @@ ResultsActivity.propTypes = {
...
 
+const mapStateToProps = state => ({
+  userRoles: state.app.user.roles,
+})
+
 const ResultsActivityWithRouter = withRouter(ResultsActivity)
 
 export { ResultsActivity }
-export default ResultsActivityWithRouter
+export default connect(mapStateToProps)(ResultsActivityWithRouter)
```
jsrobertson: Pattern tends to be to get stuff from the store in the Container above and pass it down as a prop.

PR #457, opened 2019-08-05
Revert "DT-98: Link to offenders for global search users"

petergphillips:
Reverts ministryofjustice/prisonstaffhub#456.

Not ready for links yet, need to fix one or two other bits first...


petergphillips:
only need to revert one of the commits

PR #461, opened 2019-08-06
Nn 2112 remove pay all link for future dates

Robc24:
We don't want to allow users to pay offenders for future activities so the pay all button is removed.

We want the users to be able to pay offenders who may have two activities in the list so this has been fixed.


src/ResultsActivity/ResultsActivity.js
```diff
@@ -177,6 +184,7 @@ class ResultsActivity extends Component {
       const lockedCases = attendanceInfo.filter(activity => activity.attendanceInfo.locked === true)
 
       if (
+        isAfterToday(date) ||
```
petergphillips: worth adding test for this scenario in ResultsActivity.test.js

Robc24: Good thinking.  I've added a few more tests to ResultsActivity around the date/button visibility.

PR #462, opened 2019-08-07
NN-2113 updated attendance button where other attend info exists

Robc24:
This change checks the unpaid offenders list for any offenders with 'other' attendance information already selected and updates the attend all button to say 'pay remaining...' instead of 'pay all...'

### DIFF
--- a/src/ResultsActivity/ResultsActivity.js
+++ b/src/ResultsActivity/ResultsActivity.js
@@ -198,8 +198,7 @@ class ResultsActivity extends Component {
 
     const showRemainingButton = activities => {
       const attendanceInfo = activities.filter(activity => activity.attendanceInfo)
-      if (totalPaid !== 0 || attendanceInfo.length) return true
-      return false
+      return totalPaid !== 0 || attendanceInfo.length
     }
 
     const { payingAll } = this.state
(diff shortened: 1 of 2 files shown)


src/ResultsActivity/ResultsActivity.js
```diff
@@ -196,6 +196,12 @@ class ResultsActivity extends Component {
       return showControls
     }
 
+    const showRemainingButton = activities => {
+      const attendanceInfo = activities.filter(activity => activity.attendanceInfo)
+      if (totalPaid !== 0 || attendanceInfo.length) return true
```
JamieDraperUK: could just return the result :)

src/ResultsActivity/ResultsActivity.js
```diff
@@ -196,6 +196,12 @@ class ResultsActivity extends Component {
...
+      return false
```
jsrobertson: No need for the `if`.  Return `totalPaid !== 0 || attendanceInfo.length`

src/ResultsActivity/ResultsActivity.js
Robc24: *rolley eye face*

src/ResultsActivity/ResultsActivity.js
Robc24: Aye, sorted dude.

PR #469, opened 2019-08-09
NN-1805 Fix column print widths on Missing Prisoners

jsrobertson:


jsrobertson:
> LGTM. I'll grab you to talk about the !important use.

It's needed unfortunately as this is to overwrite styles applied by `@govuk-react/table`.

PR #472, opened 2019-08-12
NN-2118 Updated selection criteria to prevent visible future lists

Robc24:


src/utils.js
```diff
@@ -41,6 +41,20 @@ const isToday = date => {
...
+const getCurrentShift = date => {
+  let timeOfDay
+
+  const afternoonSplit = 12
+  const eveningSplit = 17
+  const currentHour = moment(date).format('H')
+
+  if (currentHour < afternoonSplit) timeOfDay = 'AM'
+  if (currentHour >= afternoonSplit && currentHour < eveningSplit) timeOfDay = 'PM'
+  if (currentHour >= eveningSplit) timeOfDay = 'ED'
```
petergphillips: why do you need this `if` line?  Won't it always be true?

src/MissingPrisoners/MissingPrisonersSearch.js
```diff
@@ -39,6 +43,7 @@ const MissingPrisonersSearch = ({
 }) => (
   <Container>
     <SearchContainer>
+      {date}
```
petergphillips: did you mean to output the `date` here?

src/utils.js
petergphillips: i.e. ```
  if (currentHour < afternoonSplit) return 'AM'
  if (currentHour >= afternoonSplit && currentHour < eveningSplit) return 'PM'
  return 'ED'
```

Comment (petergphillips) [reply src/utils.js]: also has benefit of getting rid of the mutable variable :nauseated_face:

Comment (petergphillips) [reply src/utils.js]: actually even simpler
```
  if (currentHour < afternoonSplit) return 'AM'
  if (currentHour < eveningSplit) return 'PM'
  return 'ED'
```

Comment (Robc24) [reply src/utils.js]: I see what you mean.  Yeah i'm happy with that.

File: src/MissingPrisoners/MissingPrisonersSearch.js [hunk 3]
```diff
@@ -56,12 +61,17 @@ const MissingPrisonersSearch = ({
             <option key="MORNING" value="AM">
```
petergphillips: don't you need `!isTodayOrAfter` for the `am` option too, otherwise it will be shown for tomorrow won't it?

Robc24: We don't cater for future dates on this.  The date picker shows today and past dates.

PR #474, opened 2019-08-13
NN-2122 updated styled elements and added classes to handle visibility

Robc24:
- Updated visible elements to print
- removed controls from printable view
- Added date to printable view


src/MissingPrisoners/MissingPrisonersSearch.style.js
```diff
@@ -6,7 +6,7 @@ import { BORDER_COLOUR, LINK_COLOUR } from 'govuk-colours'
 
 export const Container = styled.div`
```
jsrobertson: Can just delete `Container`.

PR #478, opened 2019-08-19
Nn 2130 express template

Robc24:


package.json
```diff
@@ -155,6 +157,8 @@
     "start-feature": "PORT=3006 OAUTH_ENDPOINT_URL=http://localhost:19090/auth/ API_ENDPOINT_URL=http://localhost:18080/ NN_ENDPOINT_URL=http://localhost:20200/ API_WHEREABOUTS_ENDPOINT_URL=http://localhost:18082/ UPDATE_ATTENDANCE_PRISONS=LEI node backend/index.js",
     "build": "webpack --mode=production",
     "build:analyse": "BUNDLE_ANALYSE=true yarn build",
+    "build-css": "yarn run ./bin/build-css",
+    "css-build": "./bin/build-css",
```
andrewrlee: are both these required?

static/styles/application.css
```diff
@@ -0,0 +1,6025 @@
+.govuk-link {
+  font-family: 'GDS Transport', Arial, sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
+
+/*! Copyright (c) 2011 by Margaret Calvert & Henrik Kubel. All rights reserved. The font has been customised for exclusive use on gov.uk. This cut is not commercially available. */
```
andrewrlee: is this needed - aren't you pulling this in via node modules?

package.json
Robc24: Unlikely.  I'll check with @Nimphal though when she gets in as to whether there's a method behind the madness.

static/styles/application.css
Robc24: This is the generated CSS from the sass file.  We need to add this to the git-ignore list.

PR #481, opened 2019-08-21
NN-2119 Datepicker and print button changes

jsrobertson:
- Allow datepicker to select next 7 days by default
- Only display Print button on Activity and Housing lists for next 2 working days

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -64,6 +64,26 @@ const isAfterToday = date => {
   return daysDifference >= 0
 }
 
+const isWithinNextTwoWorkingDays = date => {
+  if (isToday(date)) return true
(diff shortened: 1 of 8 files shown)


src/utils.js
```diff
@@ -64,6 +64,26 @@ const isAfterToday = date => {
...
+  if (isSaturday) daysInAdvance += 1
+
+  return moment(date, 'DD/MM/YYYY').isBefore(
+    moment()
+      .add(daysInAdvance, 'days')
+      .endOf('day')
+  )
+}
+
 const isWithinLastYear = date => {
```
Nimphal: I really like the readability of this!

PR #485, opened 2019-08-22
NN-2119 Hide missing prisoners button for future dates and periods

jsrobertson:
Hide missing prisoners button for future dates and periods but allow any period to be selected on previous days.

### DIFF
--- a/src/Search/Search.js
+++ b/src/Search/Search.js
@@ -10,7 +10,7 @@ import styled from 'styled-components'
 import { Flag } from '../flags'
 import ValidationErrors from '../ValidationError'
 import WhereaboutsDatePicker from '../DatePickers/WhereaboutsDatePicker'
-import { isAfterToday } from '../utils'
+import { isAfterToday, getCurrentShift } from '../utils'
 
 export const MissingButtonContainer = styled('div')`
   margin-right: 5em;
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -101,6 +101,12 @@ const isWithinLastWeek = date => {
   return daysDifference >= 0
 }
 
+const isBeforeToday = date => {
+  if (isToday(date)) return false
```
petergphillips: since `isAfterToday` exists already would it be better to just do
```
const isBeforeToday = date => !(isToday(date) || isAfterToday(date))
```

src/Search/Search.js
```diff
@@ -19,6 +19,18 @@ export const MissingButtonContainer = styled('div')`
...
 
 class Search extends Component {
+  showMissingPrisonersButton = () => {
+    const { period: selectedPeriod, date } = this.props
+    const actualPeriod = getCurrentShift()
+
+    if (isBeforeToday(date)) return true
+    if (isAfterToday(date)) return false
+    if (selectedPeriod === 'AM') return true
+    if (selectedPeriod === 'PM' && (actualPeriod === 'PM' || actualPeriod === 'ED')) return true
```
petergphillips: how about:
```if (selectedPeriod === 'PM' ) return (actualPeriod === 'PM' || actualPeriod === 'ED')
```

src/utils.js
jsrobertson: Good idea, cheers

src/Search/Search.js
```diff
@@ -19,6 +19,18 @@ export const MissingButtonContainer = styled('div')`
...
+    if (selectedPeriod === 'ED' && actualPeriod === 'ED') return true
```
petergphillips: will they have always selected a `selectedPeriod`?  If so then 
```javascript
return actualPeriod === 'ED'
```

PR #489, opened 2019-08-23
NN-1857 add page limit to alert types request, add sorting

JamieDraperUK:


backend/controllers/reference-codes-service.js
```diff
@@ -15,6 +15,19 @@ const referenceCodesServiceFactory = elite2Api => {
         description: type.description,
       }))
 
+    const alphaSortOnDescription = (a, b) => {
```
petergphillips: better to do
```javascript
const alphaSortOnDescription = (a, b) => a.description.localeCompare(b.description)
```

PR #494, opened 2019-08-30
NN-1954 Close existing alert

jsrobertson:


backend/controllers/alert.js
```diff
@@ -0,0 +1,112 @@
...
+const { logError } = require('../logError')
+
+const serviceUnavailableMessage = 'Sorry, the service is unavailable'
+const getOffenderUrl = offenderNo => `${config.app.notmEndpointUrl}offenders/${offenderNo}`
+
+const alertFactory = (oauthApi, elite2Api) => {
+  const renderTemplate = (req, res, pageData) => {
+    const { alert, pageErrors, offenderDetails, ...rest } = pageData
+    const formAction =
+      offenderDetails && alert ? `/api/close-alert/${offenderDetails.bookingId}/${alert.alertId}` : undefined
```
JamieDraperUK: You could probably just use` offenderDetails && alert &&  `/api/close-alert/${offenderDetails.bookingId}/${alert.alertId}``

backend/controllers/alert.js
```diff
@@ -0,0 +1,112 @@
...
+      offenderDetails,
+      formAction,
+      alert: alert && {
+        ...alert,
+        dateCreated: formatTimestampToDate(alert.dateCreated),
+      },
+      ...rest,
+    })
+
+    if (pageErrors.length > 0) req.flash('errors', pageErrors)
```
JamieDraperUK: Does req.flash need to happen before the render?

backend/controllers/alert.js
```diff
@@ -0,0 +1,109 @@
...
+          displayName: user.name,
+          activeCaseLoad: {
+            description: activeCaseLoad && activeCaseLoad.description,
+          },
+        },
+        caseLoadId: activeCaseLoad && activeCaseLoad.caseLoadId,
+        userRoles,
+      })
+    } catch (error) {
+      logError(req.originalUrl, error)
```
Robc24: Is it worth testing this to make sure the right error is being logged?

PR #495, opened 2019-09-02
Nn 2156 circle ci css compile fix

Robc24:
- Persisted designated css folder structure to the workspace
- amended ordering of css compile and build

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,7 @@ jobs:
             - node_modules
             - build
             - build-info.json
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -39,6 +39,7 @@ jobs:
             - node_modules
             - build
             - build-info.json
+            - static/stylesheets
```
petergphillips: why do we need this line - I thought `webpack` would put it into `build` directory anyway?

PR #498, opened 2019-09-03
NN-2159 date entry update and error handling

Robc24:
Readonly date input as agreed with Andy H/Eddie/Jamie

Updated format to match GDS.

More appropriate error handling to match GDS, improved breadcrumb and link back to stats page from error page.

### DIFF
--- a/views/components/datePicker.njk
+++ b/views/components/datePicker.njk
@@ -4,5 +4,5 @@
 <label class="govuk-label" for="datepicker">
     Date
 </label>
-  <input class="govuk-input date-input" id="datepicker" name="date" type="text" value="{{date}}">
+  <input class="govuk-input date-input" id="datepicker" name="date" type="text" value="{{date}}" readonly>
 </div>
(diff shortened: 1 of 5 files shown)


views/components/datePicker.njk
```diff
@@ -4,5 +4,5 @@
 <label class="govuk-label" for="datepicker">
     Date
 </label>
-  <input class="govuk-input date-input" id="datepicker" name="date" type="text" value="{{date}}">
+  <input class="govuk-input date-input" id="datepicker" name="date" type="text" value="{{date}}" readonly>
```
petergphillips: think this is a backwards step.  Our input fields should be editable

Robc24: I see where you're coming from.  Unfortunately for us we're using components that aren't standard GDS and there are at least two other existing examples where the selection of the date is triggered by clicking in to an input that is **read only**.

Ideally we need to rework all of the datepickers and make them accessible with a button NEXT to the input giving the OPTION of the calendar but if we were to update this version and not the others we'd have another example of an inconsistent journey for the users.

To play devil's advocate, a benefit of doing it the way we are is that you are narrowing down the chance of an input error occurring by letting users input their own dates.  The datepicker is contained, with its own restrictions on what dates can be selected (future etc), so the chance of a user inputting an invalid date and the page erroring is minimal.  Unless of course you want to dick about with the querystring.  If you do want to do that then you're a boob and we'll catch you on the error page and send you back again.

Have a word with Andy/Dom if you want to discuss further *thumbs up*

PR #499, opened 2019-09-04
NN-1954 Fix broken asset path array

jsrobertson:


backend/index.js
```diff
@@ -265,7 +265,9 @@ app.get(
 )
 
 nunjucksSetup(app, path)
-;[('../node_modules/govuk-frontend/govuk/assets', '../node_modules/govuk-frontend')].forEach(dir => {
+
+const assetPaths = ['../node_modules/govuk-frontend/govuk/assets', '../node_modules/govuk-frontend']
+assetPaths.forEach(dir => {
```
petergphillips: According to https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/installing-with-npm.md you should just be able to do:
```
app.use('/assets', express.static(path.join(__dirname, '/node_modules/govuk-frontend/assets')))
```

PR #503, opened 2019-09-05
NN-2121: Refresh page instead of updating state

Nimphal:
When attending all/remaining prisoners, we are doing two loops - one loop which goes over all the offenders in the API response and calls the reducer, and then within the reducer another loop which loops over all of the offenders in state until it finds a match. When users are using the attend all functionality on pages with a lot of offenders (300), they are getting a 'something is making your browser slow' error due to the large amount of client-side processing.

Since this functionality is used on a large group of offenders at once, instead of updating state one by one, this PR proposes we simply refresh the activity list same way the page is loaded in the first place. As the attendances have been updated already through the batch API call, the returned data is equivalent to if we updated each state one by one without any client-side processing at all.

### DIFF
--- a/src/ResultsActivity/ResultsActivity.js
+++ b/src/ResultsActivity/ResultsActivity.js
@@ -85,6 +85,7 @@ class ResultsActivity extends Component {
       period,
       handlePeriodChange,
       handlePrint,
+      getActivityList,
       activityData,
       sortOrder,
       orderField,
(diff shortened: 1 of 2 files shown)


JamieDraperUK:
Changes look good but one issue might be that the date and period will reset, this might be fine though? If not to fix we will need to move period and date into query parameters.

Nimphal:
Nothing resets except potentially column ordering - the getActivityList function pulls the activity list based on the period/date in state at the point when it's called, it doesn't do anything to change it.

PR #508, opened 2019-09-09
NN-1954 add created by

JamieDraperUK:


backend/controllers/alert.js
```diff
@@ -11,14 +11,36 @@ const alertFactory = (oauthApi, elite2Api) => {
...
+      alertCode,
+      alertCodeDescription,
+      alertId,
+      alertType,
+      alertTypeDescription,
+      comment,
+      dateCreated,
+      expired,
+      addedByFirstName,
+      addedByLastName,
```
jsrobertson: Would it be better to do `const { addedByFirstName, addedByLastName, dateCreated, ...rest = alert || {{}`

PR #512, opened 2019-09-09
NN-2129 Show PEEP flag on lists

jsrobertson:
- Added PEEP alert flag to lists
- Added missing tests
- Formatted spacing between flags

### DIFF
--- a/src/AlertFlags/AlertFlags.js
+++ b/src/AlertFlags/AlertFlags.js
@@ -2,13 +2,13 @@ import React, { Fragment } from 'react'
 import PropTypes from 'prop-types'
 import './AlertFlags.scss'
 
-const Flag = ({ children, ...props }) => (
+const AlertFlag = ({ children, ...props }) => (
   <Fragment>
     <span {...props}>{children}</span>{' '}
   </Fragment>
 )
 
-Flag.propTypes = {
+AlertFlag.propTypes = {
   children: PropTypes.string.isRequired,
 }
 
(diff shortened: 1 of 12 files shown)


src/AlertFlags/AlertFlags.test.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react'
```
jsrobertson: Might change these to check for the text too.

src/AlertFlags/AlertFlags.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import './AlertFlags.scss'
+
+const AlertFlags = ({ alerts, category }) => {
+  function isShown(code) {
+    if (alerts) {
```
petergphillips: Is this correct here?  Can't be `falsey` since it would have blown up earlier with the `alerts.length`

src/AlertFlags/AlertFlags.js
```diff
@@ -0,0 +1,68 @@
...
+      return alerts.some(alert => {
+        if (alert.get) {
+          if (!alert.get('expired')) {
```
petergphillips: unnecessary nested block, would be better as
```javascript
if (!alert.get('expired') &&alert.get('alertCode') === code) return true
```

jsrobertson: As discussed, `isShown` is existing code just shows as new because of file rename.

src/AlertFlags/AlertFlags.js
```diff
@@ -1,7 +1,17 @@
-import React from 'react'
+import React, { Fragment } from 'react'
 import PropTypes from 'prop-types'
 import './AlertFlags.scss'
 
+const Flag = ({ children, ...props }) => (
+  <Fragment>
```
Nimphal: Could this be AlertFlag or something? Just thinking about confusing it with the `Flag` library component we use for feature toggles. Not a blocker as doesn't break anything, just makes searching for things a bit harder.

jsrobertson: Yeah good shout!

PR #515, opened 2019-09-10
Dt 187 add omic document skeleton page

andymarke:


backend/controllers/probationDocuments.js
```diff
@@ -0,0 +1,138 @@
...
+
+    const ensureAllowedPageAccess = userRoles => {
+      if (!userRoles.find(role => role.roleCode === 'VIEW_PROBATION_DOCUMENTS')) {
+        throw new Error('You do not have the correct role to access this page')
+      }
+    }
+
+    const getCommunityDocuments = async offenderNo => {
+      const sentenceLength = sentence => {
+        if (typeof sentence.originalLength === 'undefined' || typeof sentence.originalLengthUnits === 'undefined') {
```
JamieDraperUK: Would just checking for falsely work here? `if(!sentence.originalLength || !sentence.originalLengthUnits)` 

Just checking that you only want to display the sentence length if both length and units are present?

backend/index.js
```diff
@@ -155,6 +157,13 @@ const whereaboutsApi = whereaboutsApiFactory(
   })
 )
 
+const communityApi = communityApiFactory(
```
andrewrlee: is adding to health checks a separate story?

andymarke: No, but I'll add one in a new PR

PR #521, opened 2019-09-16
Nn 2055 upgrade eslint

Nimphal:
For https://dsdmoj.atlassian.net/browse/NN-2055

`js-yaml` was ultimately a dependency of 5 libraries we rely on:
1. `eslint` - upgraded to latest version. Some new rules added to the config relating to props spreading and missing props. Currently set to warning so we don't lose them, but the violations are so numerous it is beyond the scope of this PR to address. eslint prettier plugin also updated as a result. What could be fixed automatically has been.
2. `css-loader` - upgraded to first major version, which resolved the js-yaml issue. Attempted to upgrade further, but they have introduced some breaking changes which resulted in image urls on working for govuk-css. We should upgrade this further, but it's outside of the scope of this PR. Next version: https://github.com/webpack-contrib/css-loader/compare/v1.0.1...v2.0.0
3. `jest` - upgraded to latest version. It seems to have found way more tests than the previous version. Short of having to skip two tests as I couldn't figure out why they don't work, this was relatively easy. Now the suite reports 699 tests versus 402 before this upgrade. I suspect we weren't running all tests before.
4. `husky` - latest version, we use it very simply so that was easy
5. `lint-staged` - upgraded to the version that fixes the js-yaml and no further. They have introduced breaking changes regarding advanced options which affect us. Further upgrade is beyond the scope of this PR. Next version: https://github.com/okonet/lint-staged/releases/tag/v9.0.0

End result: only one version of js-yaml in the .lock and it's the right one!


.eslintrc
```diff
@@ -24,6 +24,10 @@
     ],
     "prettier/prettier": ["error"],
     "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
+    "react/jsx-props-no-spreading": 1,
```
jsrobertson: As discussed, understand the reason for putting the overrides in to avoid the scope getting too big.  Will look to remove them soon as time allows though.

PR #525, opened 2019-09-17
DT-188 Probation documents add to probation page for OMIC

andymarke:


views/probationDocuments.njk
```diff
@@ -43,38 +42,86 @@
...
+    </div>
+    {% if document.description %}
+      <div class="govuk-grid-column-full govuk-caption-m">
+        {{ document.description}}
+      </div>
+    {%  endif %}
+  </div>
+{% endmacro %}
+
+{% macro documentTable(documents, caption) %}
```
andrewrlee: this is a more practical way of building tables than what we've been using - we've been falling back to not using the macros as didn't want to do data transforming of of rows in the controller/custom filters (especially when we need to render buttons and things)

PR #527, opened 2019-09-18
NN-2170 added new csv download page for bulk appointments

Robc24:


Robc24:
Screenshot 2019-09-18 at 15 09 14

backend/index.js
```diff
@@ -293,6 +293,9 @@ app.get(
     probationDocumentsFactory(oauthApi, elite2Api, communityApi, oauthClientId).displayProbationDocumentsPage
   )
 )
+app.get('/need-to-upload-file', async (req, res) => {
```
Nimphal: Do we need the async here?

PR #529, opened 2019-09-19
DT-208 download probation documents

andymarke:


backend/api/oauthEnabledClient.js
```diff
@@ -122,9 +122,36 @@ const factory = ({ baseUrl, timeout }) => {
         })
     })
 
+  const pipe = (context, path, pipeTo) => {
+    const url = remoteUrl + path
```
JamieDraperUK: Might be able to use url.resolve here.

PR #551, opened 2019-09-26
NN-2171 add conditional reveal

JamieDraperUK:


backend/controllers/addAppointmentDetails.js
```diff
@@ -23,6 +23,67 @@ const setSelected = (value, items) =>
...
+  return time && time.hour(Number(hours)).minutes(Number(minutes))
+}
+
+const extractHoursMinutes = dateTime => {
+  const instant = moment(dateTime, DATE_TIME_FORMAT_SPEC)
+  const hours = instant.hours()
+  const minutes = instant.minutes()
+
+  return {
+    hours: hours <= 9 ? `0${hours}` : hours,
```
petergphillips: this looks complicated and can return either a string or number (string if < 10).  You've got an `instant` anyway so think it would be better to do:
```javascript
hours: instant.format('HH')
minutes: instant.format('mm')
```

backend/controllers/addAppointmentDetails.js
```diff
@@ -41,11 +102,26 @@ const addAppointmentDetailsFactory = (bulkAppointmentService, oauthApi, logError
...
       res.render('addAppointmentDetails.njk', {
         title: 'Add appointment details',
         appointmentTypes,
         locations,
         homeUrl: notmEndpointUrl,
+        appointmentType,
+        location,
+        comments,
+        date,
+        startTimeHours: startTime && String(startTimeHoursMinutes.hours),
```
petergphillips: you're already testing to see if `startTime` set above, so this is a duplicate.  If you use `format` instead then casting to `String` here won't be required.  Maybe `extractHoursMinutes` could cope with time not being set and then you could just use spread operator here for hours and minutes

JamieDraperUK: You're right, let me clean the mapping code up a bit.

backend/controllers/addAppointmentDetails.js
```diff
@@ -23,6 +23,67 @@ const setSelected = (value, items) =>
...
+
+    if (isToday && startTimeDuration.asMinutes() > 1)
+      errors.push({ text: 'Select a start time that is not in the past', href: '#start-time-hours' })
+
+    if (endTime && endTimeDuration.asMinutes() > 1) {
+      errors.push({ text: 'Select an end time that is not in the past', href: '#end-time-hours' })
+    }
+  }
+
+  if (comments && comments.length > 3600)
```
petergphillips: out of curiosity why are you doing this manually rather than using the https://design-system.service.gov.uk/components/character-count/ component (instead of textarea)?

backend/controllers/addAppointmentDetails.js
```diff
@@ -23,6 +23,67 @@ const setSelected = (value, items) =>
         : item
   )
 
+const buildDateTime = ({ date, hours, minutes }) => {
+  const time = date && hours && minutes && moment(date, DAY_MONTH_YEAR)
+  return time && time.hour(Number(hours)).minutes(Number(minutes))
+}
+
+const extractHoursMinutes = dateTime => {
+  const instant = moment(dateTime, DATE_TIME_FORMAT_SPEC)
```
petergphillips: (see below comments) so think here you could cope with not being set by doing something like:
```javascript
if (!dateTime) return { hours: '', minutes: ''}
```

backend/controllers/addAppointmentDetails.js
JamieDraperUK: Andy H didn't want to use that component due to the fear of creating too much noise.

petergphillips: :astonished:

PR #569, opened 2019-10-03
NN-2186 add recurring info to the upload page

JamieDraperUK:


backend/controllers/addAppointmentDetails.js
```diff
@@ -2,7 +2,7 @@ const moment = require('moment')
 const config = require('../config')
 const { serviceUnavailableMessage } = require('../common-messages')
 const { DATE_TIME_FORMAT_SPEC, DAY_MONTH_YEAR, buildDateTime } = require('../../src/dateHelpers')
-
+const { properCaseName } = require('../utils')
```
jsrobertson: Not a show stopper but wondering if this is better off being called `capitalise` now it's being used for things other than names.

JamieDraperUK: That's a good spot!

JamieDraperUK: Will change it on the next PR

jsrobertson: Oh, there is already a `capitalize` in utils

PR #570, opened 2019-10-03
NN-2186 move appointment details to component

JamieDraperUK:


views/components/appointment-details-summary/appointmentDetailsSummary.njk
```diff
@@ -0,0 +1,69 @@
+
```
jsrobertson: Nice one for doing this.

PR #574, opened 2019-10-03
migrate deployment to cp

mattops:


.circleci/config.yml
```diff
@@ -80,74 +135,86 @@ jobs:
...
     jobs:
-      - build
+      - build:
+          filters:
+            tags:
+              ignore: /.*/
       - integration_tests:
           requires:
-            - build
+          - build
```
petergphillips: I thought indenting matters - looks to be out here?

.circleci/config.yml
```diff
@@ -80,74 +135,86 @@ jobs:
...
       - build_docker:
           requires:
             - integration_tests
           filters:
             branches:
-              only:
-                - master
-      - deploy_release_eb:
+              only: master
```
petergphillips: indenting looks wrong here?

PR #583, opened 2019-10-08
NN-2176 Appointment clashes page

jsrobertson:


backend/controllers/bulkAppointmentsClashes.js
```diff
@@ -0,0 +1,115 @@
...
+
+const bulkAppointmentsClashesFactory = (elite2Api, logError) => {
+  const renderTemplate = (req, res, pageData) => {
+    const { appointmentDetails, prisonersWithClashes } = pageData
+
+    res.render('appointmentsClashes.njk', { appointmentDetails, prisonersWithClashes })
+  }
+
+  const renderError = (req, res, error) => {
+    if (error) logError(req.originalUrl, error, 'Sorry, the service is unavailable')
```
JamieDraperUK: Probs should use the shared messages module.

backend/controllers/bulkAppointmentsClashes.js
```diff
@@ -0,0 +1,115 @@
...
+      } = req.session
+
+      const eventsByOffenderNo = await getOtherEvents(req, res, {
+        offenderNumbers: prisonersListed.map(prisoner => prisoner.offenderNo),
+        date: switchDateFormat(date),
+        agencyId: req.session.userDetails.activeCaseLoadId,
+      }).then(events =>
+        events.reduce(
+          (offenders, event) =>
+            Object.assign(offenders, { [event.offenderNo]: (offenders[event.offenderNo] || []).concat(event) }),
```
JamieDraperUK: Why not just use the spread operator?

backend/controllers/bulkAppointmentsClashes.js
```diff
@@ -0,0 +1,115 @@
...
+        startTime: startTime || buildDateTime({ date, hours: 23, minutes: 59 }).format(DATE_TIME_FORMAT_SPEC),
+        endTime,
+      },
+      appointments: remainingPrisoners.map(prisoner => ({
+        bookingId: prisoner.bookingId,
+        startTime: prisoner.startTime,
+        endTime: prisoner.endTime,
+      })),
+      repeat:
+        recurring === 'yes'
```
JamieDraperUK: Could we just use:
`&& {
              repeatPeriod: repeats,
              count: Number(times),
 } `

Because the falsy value will be undefined.

backend/tests/bulkAppointmentsClashes.test.js
```diff
@@ -0,0 +1,376 @@
...
+              endTimeMinutes: '15',
+              startTime: '2019-10-08T18:50:00',
+              endTime: '2019-10-08T19:15:00',
+            },
+          ],
+        }
+      })
+
+      it('submit the data and redirect to the appointments added page', async () => {
+        jest.spyOn(Date, 'now').mockImplementation(() => 1569481200000) // Thursday 2019-09-26T07:00:00.000Z
```
JamieDraperUK: Is this needed? Not sure we're doing any new instances of date times/moment() in the controller class? Sorry if I've missed it.

jsrobertson: Yeah, when we do the supply a default start time when there are individual offender start times.

backend/controllers/bulkAppointmentsClashes.js
JamieDraperUK: Ignore ^^^ the falsy value will be false.

PR #591, opened 2019-10-10
NN-2182 Printed movement authorisation slips

jsrobertson:
image


backend/controllers/bulkAppointmentsClashes.js
```diff
@@ -67,6 +67,7 @@ const bulkAppointmentsClashesFactory = (elite2Api, logError) => {
 
   const post = async (req, res) => {
     const {
+      data,
```
andrewrlee: didn't know you could refer to the same thing more than once when destructuring 🤓

PR #592, opened 2019-10-10
Nn 2207 alpha occurrence input

Robc24:
- removed 'number' type from appointment occurrences input


jsrobertson:
I think `novalidate` on the `<form />` would handle this

PR #595, opened 2019-10-11
NN-2212 Redirect to no appointments added when no prisoners remain after removing clashes

jsrobertson:


backend/controllers/bulkAppointmentsUpload.js
```diff
@@ -60,7 +60,7 @@ const bulkAppointmentsUploadFactory = (csvParserService, offenderLoader, logErro
           const offenderNosNotFound = getNonExistingOffenderNumbers(offendersFromCsv, prisonerList)
 
           if (offenderNosNotFound.length === fileContent.length) {
-            res.redirect('/bulk-appointments/no-appointments-added')
+            res.redirect('/bulk-appointments/no-appointments-added?reason=offendersNotFound')
             res.end()
```
JamieDraperUK: Prob best to return after this?

JamieDraperUK: What's the res.end() used for?

jsrobertson: No idea, I'm just changing it now to return res.redirect...

PR #619, opened 2019-10-16
NN-2030: Make sure correct event is shown for absence, update page heading

Nimphal:
Previous we were returning the first event, which wasn't necessarily the one the absence was for. This resolves and updates the page heading.

### DIFF
--- a/backend/controllers/attendanceStatistics.js
+++ b/backend/controllers/attendanceStatistics.js
@@ -153,7 +153,9 @@ const attendanceStatisticsFactory = (oauthApi, elite2Api, whereaboutsApi, logErr
       const absences = attendances.filter(absence => absence.absentReason === reason)
 
       const offenderData = absences.map(absence => {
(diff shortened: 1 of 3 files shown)


backend/controllers/attendanceStatistics.js
```diff
@@ -153,7 +153,9 @@ const attendanceStatisticsFactory = (oauthApi, elite2Api, whereaboutsApi, logErr
       const absences = attendances.filter(absence => absence.absentReason === reason)
 
       const offenderData = absences.map(absence => {
-        const offenderActivity = activities.find(activity => activity.bookingId === absence.bookingId)
+        const offenderActivity = activities.find(
+          activity => activity.bookingId === absence.bookingId && activity.eventId === absence.eventId
```
andrewrlee: guess your test data doesn't cover this specific case?

Nimphal: No, I'll probably look at updating that in a later PR.

PR #620, opened 2019-10-16
NN-2182 Allow large amount of movement slips to be printed

jsrobertson:


backend/controllers/bulkAppointmentsSlipsController.js
```diff
@@ -22,13 +22,23 @@ module.exports = ({ elite2Api, logError }) => async (req, res) => {
...
     const offenderNumbers = prisonersListed.map(prisoner => prisoner.offenderNo)
-    const offenderSummaries = await elite2Api.getOffenderSummaries(res.locals, offenderNumbers)
+    const chunkedOffenderNumbers = chunkArray(offenderNumbers, 100)
+
+    const offenderSummaryApiCalls = chunkedOffenderNumbers.map(offendersChunk => ({
+      getOffenderSummaries: elite2Api.getOffenderSummaries,
+      offenders: offendersChunk,
+    }))
+
+    const offenderSummaries = await Promise.all(
```
JamieDraperUK: Would you be better of doing:
`(await Promise.all(stuff)).reduce(`

jsrobertson: That won't work.  Think it makes more sense for the Promise.all to be chained with `.then` which contains the responses so they can then be reduced them into one array.

jsrobertson: As discussed, didn't see you'd wrapped it in `()`

PR #621, opened 2019-10-16
NN-2218 Handle Prison number CSV header

jsrobertson:


backend/csv-parser.js
```diff
@@ -20,7 +20,7 @@ const csvParserService = ({ fs, isBinaryFileSync }) => {
       let record
       // eslint-disable-next-line no-cond-assign
       while ((record = this.read())) {
-        output.push(record)
+        if (record[0] !== 'Prison number') output.push(record)
```
JamieDraperUK: Not sure this is the right place for this logic as this parser is generric. Maybe a better place would be in bulkAppointmentsUpload.js?

PR #622, opened 2019-10-16
NN-2220 removed conditions which prevent session update

Robc24:
- Updates to session were conditional on there being missing/prisoners in the list.  This isn't the way it should work as an empty list can be passed through.

### DIFF
--- a/backend/controllers/bulkAppointmentsUpload.js
+++ b/backend/controllers/bulkAppointmentsUpload.js
@@ -63,12 +63,9 @@ const bulkAppointmentsUploadFactory = (csvParserService, offenderLoader, logErro
             return res.redirect('/bulk-appointments/no-appointments-added?reason=offendersNotFound')
           }
 
-          if (prisonerList && prisonerList.length) {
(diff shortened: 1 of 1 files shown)


backend/controllers/bulkAppointmentsUpload.js
```diff
@@ -63,12 +63,9 @@ const bulkAppointmentsUploadFactory = (csvParserService, offenderLoader, logErro
...
           }
 
-          if (prisonerList && prisonerList.length) {
-            req.session.data.prisonersListed = prisonerList
-          }
-          if (offenderNosNotFound && offenderNosNotFound.length) {
-            req.session.data.prisonersNotFound = offenderNosNotFound
-          }
+          req.session.data.prisonersListed = prisonerList
+
```
Nimphal: Could these be undefined instead of empty lists? Is it safer to do:
```
req.session.data.prisonersListed = prisonerList || []
``` 

and same for the below?

Robc24: As discussed :)

PR #649, opened 2019-10-24
NN-2238 removed old bulk appointments functionality

Robc24:
- removed old bulk appts functionality
- tweaked bulk appts added redirect as it doesn't need to clear the session data anymore


backend/controllers/bulkAppointmentsFormRedirect.js
```diff
@@ -1,5 +1,3 @@
 module.exports = (req, res) => {
```
JamieDraperUK: This is a strange module. Where is it used?

backend/tests/bulkAppointmentsFormRedirect.test.js
```diff
@@ -6,25 +6,11 @@ const mockRes = { redirect: jest.fn() }
 describe('add appointment redirect', () => {
```
JamieDraperUK: Is this part of the new bulk appointment work?

backend/controllers/bulkAppointmentsFormRedirect.js
Robc24: It was used previously to clear out the session data and do the redirect to the add appointments.  Looking at it now, I don't think it's needed.  We can just redirect to the form directly as the session has already been cleared out.  What do you reckon?

JamieDraperUK: Yeah, my preference is to remove entirely.

backend/tests/bulkAppointmentsFormRedirect.test.js
Robc24: It was put together to cater for the session clear and redirect.  I removed the session part but we've no need for any of it.  About to remove the whole thing.

PR #656, opened 2019-11-01
Correct Licence URL

Mjwillis:


helm_deploy/values-preprod.yaml
```diff
@@ -23,7 +23,7 @@ env:
   API_WHEREABOUTS_ENDPOINT_URL: https://whereabouts-api-preprod.service.justice.gov.uk/
   IEP_CHANGE_LINK_ENABLED: "true"
   UPDATE_ATTENDANCE_PRISONS: "MDI,HMI,NHI,PVI,WWI,BXI,STI,LEI,HLI,GHI,HHI,WLI,WCI,BLI,HEI,LLI,LNI,UKI,UPI,CFI,LHI,PVI,KMI,RNI,GTI,EYI,TCI,WYI,SWI,NWI,MRI,DTI,KVI,DMI,FKI,SKI,WNI"
-  LICENCES_URL: https://licences-preprod.service.hmpps.dsd.io/
+  LICENCES_URL: https://licences-preprod.prison.service.justice.gov.uk
```
brightonsbox: Do you need to keep the trailing / at the end?

Mjwillis: good point, not sure but will change to match

PR #664, opened 2019-11-06
NN-2127 Staying or leaving wing filter

jsrobertson:


backend/controllers/houseblockList.js
```diff
@@ -26,6 +26,9 @@ const shouldPromoteToMainActivity = (offender, newActivity) => {
   return Boolean(safeTimeCompare(mainActivity.startTime, newActivity.startTime))
 }
 
+const isStayingOnWing = locationCode =>
+  locationCode === 'WOW' || locationCode === 'STAYONWING' || locationCode === 'UNEMPLOYED' || locationCode === 'RETIRED'
```
JamieDraperUK: Might look cleaner doing, `locationCode => ['WOW','STAYONWING','UNEMPLOYED','RETIRED'].includes(locationCode)`

backend/tests/houseblockList.test.js
```diff
@@ -671,4 +673,337 @@ describe('Houseblock list controller', () => {
...
+      ])
+    })
+
+    it('should return only offenders staying on the wing', async () => {
+      elite2Api.getHouseblockList.mockImplementationOnce(() => responseWithOneLeavingWing)
+
+      const response = await houseblockList({}, 'LEI', 'Houseblock 1', '15/10/2017', 'ED', 'staying')
+
+      expect(response.length).toEqual(4)
+      expect(response).toEqual([
```
JamieDraperUK: There's a lot of test data here, I wonder if it'll be easier to reason about if `expect.objectContaining` was used, to validate part of the data structure?  That way the dataset would be a bit clearer.

PR #683, opened 2019-11-18
Nn 2246 caseload switcher

Nimphal:
For https://dsdmoj.atlassian.net/browse/NN-2246

This replaces the current React caseload switching mechanism with a new page, linked to from both React and non react apps, implementing a caseload change form. The change is propagated in both the React and non-react sides of the application.


backend/controllers/changeCaseload.js
```diff
@@ -0,0 +1,62 @@
...
+    // Don't call API if data is missing
+    if (caseLoadId) {
+      await elite2Api.setActiveCaseload(res.locals, { caseLoadId })
+
+      if (req.session && req.session.userDetails) req.session.userDetails.activeCaseLoadId = caseLoadId
+      if (req.session && req.session.data) req.session.data = null
+
+      res.redirect(config.app.notmEndpointUrl)
+    } else {
+      logError(req.originalUrl, 'Caseload ID is missing')
```
JamieDraperUK: Would it be better here to use GDS style errors? Where you can click the error and it would focus the select box. Have a look at either stats or bulk appointments.

JamieDraperUK: Actually, ignore that. You'd only get this error if you managed to clear out the format values.

Nimphal: 👍 yeah, the form doesn't have an empty state for the user, just crossing my t's here :D

views/changeCaseload.njk
```diff
@@ -0,0 +1,36 @@
...
+<h1 class="govuk-heading-xl">{{title}}</h1>
+<form method="POST" id="changeCaseload">
+  {{ govukSelect({
+          id: "changeCaseloadSelect",
+          name: "caseLoadId",
+          label: {
+              text: "Select active case load"
+          },
+          items: options | setSelected(caseLoadId),
+          errorMessage: errors | findError('changeCaseload')
```
JamieDraperUK: Not sure errorMessage is needed here.

PR #688, opened 2019-11-25
NN-2216 Added prisoners not found page to BA journey

jsrobertson:
Also re-added the integration tests which were deleted in error as part of NN-2238 😮 

image

### DIFF
--- a/views/prisonersNotFound.njk
+++ b/views/prisonersNotFound.njk
@@ -1,8 +1,7 @@
 {% extends "./partials/layout.njk" %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 {% from "govuk/components/back-link/macro.njk" import govukBackLink %}
-{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
-{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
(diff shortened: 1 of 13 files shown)


prisonstaffhub-specs/src/test/resources/offenders-for-appointments.csv
```diff
@@ -1,3 +1,2 @@
-offenderNo
+Prison number
```
JamieDraperUK: Did you mean to check this in?

JamieDraperUK: Ignore that, just remembered it's for the integration tests :)

views/prisonersNotFound.njk
```diff
@@ -0,0 +1,69 @@
...
+
+    <p class="govuk-body">
+      Some prisoners could not be found. This might be because the prison number and the name do not match, or the prison number does not exist in this location.
+    </p>
+
+    <p class="govuk-body">
+      You should check these numbers. No appointments will be created for these prison numbers.
+    </p>
+
+    <table class="govuk-table" data-qa="prisoners-not-found">
```
JamieDraperUK: Is it worth using the Table macro?

PR #690, opened 2019-11-27
NN-2297 Display duplicate numbers from CSV on new invalid numbers page

jsrobertson:


backend/controllers/bulkAppointmentsUpload.js
```diff
@@ -54,8 +54,14 @@ const bulkAppointmentsUploadFactory = (csvParserService, offenderLoader, logErro
             activeCaseLoadId
           )
 
-          const removeDuplicates = array => [...new Set(array)]
-          const offendersFromCsv = removeDuplicates(fileContentWithNoHeader.map(row => row[0]))
+          const [duplicatedPrisoners, nonDuplicatedPrisoners] = fileContentWithNoHeader.reduce(
+            (separatedList, offender) => {
+              const [offenderNumber] = offender
+              separatedList[JSON.stringify(separatedList[1]).includes(offenderNumber) ? 0 : 1].push(offenderNumber)
```
JamieDraperUK: Is the use of JSON.stringify here to get access to the includes method?

JamieDraperUK: I wonder if it might be clearer if we split this out a bit? Feels a bit hard to read.

jsrobertson: Had left the stringify in by mistake after trying something else out before I did this as `[['G1683VN'], ['G4803UT']].includes(['G1683VN']) === false`

PR #702, opened 2019-12-10
NN-2235 include suspended activity lists

JamieDraperUK:
Screenshot 2019-12-10 at 14 00 49


src/ResultsActivity/ResultsActivity.js
```diff
@@ -339,6 +339,16 @@ class ResultsActivity extends Component {
 
     const renderMainEvent = event => {
       const mainEventDescription = `${getHoursMinutes(event.startTime)} - ${getMainEventDescription(event)}`
+      if (event.suspended) {
+        return (
+          <td className="row-gutters">
```
jsrobertson: Does this need a corresponding `<th />`?

PR #710, opened 2019-12-17
NN-2344 show recurring on appointment confirmation page

JamieDraperUK:


Nimphal:
You're committing a package-lock.json, is that on purpose? Nothing wrong with it, but it might get out of date with yarn lock unless we run npm as well

PR #716, opened 2019-12-18
NN-2338 pre post appointments

JamieDraperUK:


backend/controllers/prepostAppoinments.js
```diff
@@ -0,0 +1,226 @@
...
+            endTime,
+            comment,
+            recurring,
+            times,
+            repeats,
+          }),
+        })
+      }
+
+      await createAppointment(res.locals, appointmentDetails)
```
JamieDraperUK: These don't fail together, we may need to make use of the bulk appointments API, however a special role is required to use it. We should revisit this later.

backend/controllers/prepostAppoinments.js
```diff
@@ -0,0 +1,226 @@
...
+
+  if (postAppointment === 'yes' && !postAppointmentLocation)
+    errors.push({ text: 'Select a room', href: '#postAppointmentLocation' })
+
+  return errors
+}
+
+const getLinks = offenderNo => ({
+  postAppointments: `/offenders/${offenderNo}/prepost-appointments`,
+  prisonerProfile: `${dpsUrl}offenders/${offenderNo}`,
```
brightonsbox: Does the dpsUrl always end with a '/'?

JamieDraperUK: It does :)

PR #717, opened 2019-12-18
Nn 2347 make end time non optional vlb

Nimphal:


views/addAppointment/addAppointment.njk
```diff
@@ -185,7 +185,7 @@
               }
           }) }}
 
-          {{ govukButton({ text: "Add appointment", type: "submit" }) }}
+          {{ govukButton({ text: "Continue", type: "submit" }) }}
```
JamieDraperUK: Should be save and continue

PR #736, opened 2020-01-09
Nn 2342 handle existing attendance error

Nimphal:
The whereabouts API now returns a 409 conflict. This is to show a cleaner message to the user and also let them refresh the page. Includes a refresh function for the activity list view.


src/ResultsActivity/ResultsActivityContainer.js
```diff
@@ -123,6 +125,10 @@ class ResultsActivityContainer extends Component {
     )
   }
 
+  update() {
+    this.getActivityList()
+  }
```
Nimphal: Put this in its own update to keep consistency with the other pages, plus we can add sorting behaviour on refresh here in the future.

PR #737, opened 2020-01-10
NN-2366: Reset errors on component unmount in activity list and unaccounted for

Nimphal:


src/PrisonersUnaccountedFor/PrisonersUnaccountedForContainer.js
```diff
@@ -77,6 +77,8 @@ function PrisonersUnaccountedForContainer({
 
       getPrisonersUnaccountedFor()
       setReloadPage(false)
+
+      return resetErrorDispatch
```
Nimphal: Apparently the return from useEffect is what is used on unmount 🤷‍♀

PR #747, opened 2020-01-16
NN-2375: Add custom video link confirmation page

Nimphal:


backend/controllers/appointmentsService.js
```diff
@@ -32,6 +37,12 @@ const toAppointmentDetailsSummary = ({
     recurring: properCaseName(recurring),
     ...recurringInformation,
   }
+
+  if (isVideoLinkBooking(appointmentType)) {
+    ;['appointmentType', 'recurring'].forEach(e => delete appointmentInfo[e])
+  }
```
Nimphal: It made a bit more sense to customise for VLB, rather than have a check for if the appointment isn't VLB.

PR #755, opened 2020-01-21
Nn 2343 install notify

Nimphal:


backend/config.js
```diff
@@ -51,4 +51,7 @@ module.exports = {
   updateAttendancePrisons: process.env.UPDATE_ATTENDANCE_PRISONS
     ? process.env.UPDATE_ATTENDANCE_PRISONS.split(',').map(agency => agency.trim())
     : [],
+  notifications: {
+    notifyKey: process.env.NOTIFY_API_KEY || '',
+  },
```
Nimphal: I've added the secret to dev, but we'd need the live secret to add it to prod.

Nimphal:
You can generate your own keys for local dev and pick which option you want in terms of sending emails.

PR #757, opened 2020-01-22
NN-2390 pull agency from url

JamieDraperUK:


views/addAppointment/addCourtAppointment.njk
```diff
@@ -72,14 +72,6 @@
                     attributes: {'data-disable-past-date':'true', 'data-add-appointment-date': 'true'}
                 }) }}
 
-                <div class="govuk-grid-row">
```
JamieDraperUK: Going to be replaced in later commits.

PR #762, opened 2020-01-27
NN-2396 Updated video link booking journey for court users

jsrobertson:


backend/controllers/appointments/selectCourtAppointmentRooms.js
```diff
@@ -0,0 +1,272 @@
+const moment = require('moment')
```
jsrobertson: Adapted along with tests from `prePostAppointments.js`.

PR #764, opened 2020-01-29
Nn 2365 rename iep files

Nimphal:
Renames a bunch of files and routes relating to IEP name change to Incentive level. Does not touch variable and function names as part of this PR.


helm_deploy/prison-staff-hub/templates/_envs.tpl
```diff
@@ -79,9 +79,6 @@ env:
   - name: API_WHEREABOUTS_ENDPOINT_URL
     value: {{ .Values.env.API_WHEREABOUTS_ENDPOINT_URL | quote }}
 
-  - name: IEP_CHANGE_LINK_ENABLED
-    value: {{ .Values.env.IEP_CHANGE_LINK_ENABLED | quote }}
-
```
Nimphal: This feature is no longer behind a flag, this variable is not being used.

src/routePaths.js
```diff
@@ -2,7 +2,8 @@ export default {
   establishmentRoll: '/establishment-roll',
   inToday: '/establishment-roll/in-today',
   iepHistory: '/offenders/:offenderNo/iep-details',
```
Nimphal: Keeping old URL very briefly because new-nomis-ui needs changing to point to the new one as well.

src/routes.js
```diff
@@ -15,7 +15,11 @@ export default [
   { path: '/offenders', breadcrumb: null },
   { path: '/offenders/:offenderNo', breadcrumb: OffenderBreadcrumb, renderDirectly: true },
   { path: '/offenders/:offenderNo/iep-details', breadcrumb: 'Incentive details' },
```
Nimphal: Same for this url, just making sure things don't break in case we don't quite deploy everything in order.

PR #767, opened 2020-01-30
Nn 2400 email confirmation court user

Nimphal:
Adds email confirmation to the adding a videolink booking journey. Nobody that's not on the team will be getting emailed yet, this is preliminary work.


backend/controllers/appointments/prepostAppoinments.js
```diff
@@ -6,6 +6,7 @@ const {
 
 const { serviceUnavailableMessage } = require('../../common-messages')
 const { toAppointmentDetailsSummary } = require('./appointmentsService')
+const { notifyClient } = require('../../shared/notifyClient')
```
JamieDraperUK: Would be nice to see notifyClient as an injectable dependency like the others `({ elite2Api, oauthApi, appointmentsService, existingEventsService, logError }) `

Nimphal: I was wondering between that and importing it where needed. It does mean we're initialising it multiple times this way, which I've just about come around to thinking about :D

backend/controllers/appointments/prepostAppoinments.js
```diff
@@ -292,6 +299,34 @@ const prepostAppointmentsFactory = ({ elite2Api, appointmentsService, existingEv
         ...prepostAppointments,
       })
 
+      const userEmailData = await oauthApi.userEmail(res.locals, username)
+
+      const templateId = '391bb0e0-89b3-4aef-b11e-c6550b71fee8'
```
JamieDraperUK: Is it worth moving this into config.js?

Nimphal: I can't decide yet. I am kind of hoping we end up with something abstracted when we settle on how and when we send emails to people. Left it almost on purpose to look ugly. Happy to move to config though, easily done.

PR #768, opened 2020-01-31
Nn 2397 court appointment availability part2

JamieDraperUK:
Checks to see if the appointment locations haven't been booked for the requested time slot. 

Future work is required around checking to see if the rooms are still available just before creating the appointments. Also, we need some discussion around what to do with failures when it comes to either emails failing or the appointment creation.


prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/specs/AddCourtAppointmentSpecification.groovy
```diff
@@ -59,9 +58,9 @@ class AddCourtAppointmentSpecification extends BrowserReportingSpec {
         at SelectCourtAppointmentRoomsPage
 
         and: "I fill out the form"
-        selectRoomsForm.preAppointmentLocation = 1
-        selectRoomsForm.mainAppointmentLocation = 2
-        selectRoomsForm.postAppointmentLocation = 3
+        selectRoomsForm.selectPreAppointmentLocation = 1
```
JamieDraperUK: There is some Javascript that's always loaded that's used by a different page that has some pre and post elements on it, however there are change events hooked into the ids listed above. So for now I've just renamed them to avoid clashes. Longer time we either need to reduce the Javascript, reduce the scope with a named class, or load the js for specific pages.

jsrobertson: We do classes prefixed with `js-` for the other appointment stuff so makes sense to carry on doing that.

PR #779, opened 2020-02-14
DT-526 Do not retrieve offender activities if there are no cells in t…

mikehalmamoj:
…he wing

### DIFF
--- a/backend/tests/houseblockList.test.js
+++ b/backend/tests/houseblockList.test.js
@@ -274,7 +274,7 @@ describe('Houseblock list controller', () => {
 
     expect(whereaboutsApi.getAgencyGroupLocations.mock.calls.length).toBe(1)
     expect(elite2Api.getHouseblockList.mock.calls.length).toBe(0)
-    expect(response).toStrictEqual({})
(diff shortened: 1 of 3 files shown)


backend/tests/houseblockList.test.js
```diff
@@ -197,7 +197,7 @@ describe('Houseblock list controller', () => {
       triggersIEPWarning: ['UnacceptableAbsence', 'Refused'],
     })
     whereaboutsApi.getAgencyGroupLocations = jest.fn()
-    whereaboutsApi.getAgencyGroupLocations.mockReturnValue([])
+    whereaboutsApi.getAgencyGroupLocations.mockReturnValue([1])
```
petergphillips: not sure this looks correct - don't you need something like
```
[ { locationId: 1 } ]
```
instead since the code maps to `locationId` ?

petergphillips: guess it doesn't matter though as only test data

mikehalmamoj: Yes you're right.  The result gets thrown away by the test though... so I might just drop this in the next PR.

PR #784, opened 2020-02-19
NN-2439 Get list of courts from new endpoint

jsrobertson:


backend/controllers/appointments/selectCourtAppointmentCourt.js
```diff
@@ -14,13 +14,20 @@ const unpackAppointmentDetails = req => {
...
+      const { courtLocations } = await whereaboutsApi.getCourtLocations(context)
+
+      const courts = courtLocations.reduce((courtList, court) => {
+        const key = court.replace(/\W+/g, '-').toLowerCase()
+        return { ...courtList, [key]: court }
+      }, {})
+
+      return courts
+    } catch (error) {
+      throw new Error(error)
```
JamieDraperUK: Is there any need to catch the Error?

PR #788, opened 2020-02-24
NN-2300 no availability part 1

JamieDraperUK:
- Check availability of bookable locations throughout the day


backend/services/availableSlotsService.js
```diff
@@ -0,0 +1,196 @@
...
+      .hour(Number(endOfDay))
+      .minute(0)
+      .seconds(0)
+      .millisecond(0)
+
+    const duration = moment.duration(endTime.diff(startTime))
+    const totalMinutes = duration.asMinutes()
+    const numberOfSlots = Math.floor(totalMinutes / byMinutes)
+
+    return [...Array(numberOfSlots).keys()]
```
Nimphal: Nice :+1:

backend/controllers/appointments/addCourtAppointment.js
```diff
@@ -143,11 +130,58 @@ const addCourtAppointmentsFactory = (existingEventsService, elite2Api, logError)
...
+        const availableRooms = await availableSlotsService.getAvailableRooms(res.locals, {
+          startTime: start,
+          endTime: end,
+          agencyId,
+        })
 
-      if (!mainLocations.length || !preAppointmentLocationsValid || !postAppointmentLocationsValid) {
-        return res.send('No rooms available')
+        const atLeastTwoRoomsNeeded = Boolean(preAppointmentRequired === 'yes' || postAppointmentRequired === 'yes')
+        const minRooms = atLeastTwoRoomsNeeded ? 2 : 1
```
andrewrlee: I don't really understand the domain - why do you need 2 rooms if doing a pre or post appointment? Suggests both rooms would be used at the same time.

JamieDraperUK: You need at least 2 because of the following:

- Pre and post meeting can use the same room
- Pre and post meeting can't use the same location as the main one

JamieDraperUK: Also worth noting that the pre and post appointments are 25 min each and the available slots service brings back locations that are available from the start time of the _pre_ and the end time of the _post_. With that it shouldn't be possible for the pre to overlap with the main appointment.

PR #793, opened 2020-02-25
NN-2404 fix build time function

JamieDraperUK:
Fixes a classic JavaScript falsely bug in the buildDateTime function.


src/dateHelpers.js
```diff
@@ -10,7 +10,11 @@ const DayMonthYear = dateTime => moment(dateTime, DATE_TIME_FORMAT_SPEC).format(
 const Time = dateTime => moment(dateTime, DATE_TIME_FORMAT_SPEC).format('HH:mm')
 
 const buildDateTime = ({ date, hours, minutes }) => {
-  const time = date && hours && minutes && moment(date, DAY_MONTH_YEAR)
+  const time =
+    date &&
+    Number.isSafeInteger(Number.parseInt(hours, 10)) &&
```
JamieDraperUK: If either the hour or minutes is zero then it'll return falsely rather than a valid date time.

PR #802, opened 2020-02-26
NN-2421: Upgrade to node 12

Nimphal:


src/ResultsActivity/activityResultsSorter.test.js
```diff
@@ -84,11 +84,11 @@ describe('Activity results sorter tests', () => {
       expect(data).toEqual([
         { eventType: 'PRISON_ACT', comment: 'A' },
         { event: 'PA', comment: 'B' },
+        { eventDescription: 'C' },
```
Nimphal: The sorting algorithm has changed between node 10 and node 12, resulting in slightly altered behaviour. Browsers have their own sorting behaviour implementation, so the users shouldn't see a difference and from this tests it's clear that it's a very small change regarding possible missing data points, which would not occur in line.

PR #809, opened 2020-02-28
NN-2444 Add final room availability check on room selection page

jsrobertson:


backend/controllers/appointments/selectCourtAppointmentRooms.js
```diff
@@ -286,6 +287,29 @@ const selectCourtAppointmentRoomsFactory = ({
...
 
+      const availableRooms = await availableSlotsService.getAvailableRooms(res.locals, {
+        startTime,
+        endTime,
+        agencyId,
+      })
+
+      const isRoomStillAvailable = selectedRoom => availableRooms.some(room => room.value === Number(selectedRoom))
+
+      const preLocationAvailableOrNotRequired = selectPreAppointmentLocation
```
JamieDraperUK: Could this be simplified to `
 const preLocationAvailableOrNotRequired = Boolean(selectPreAppointmentLocation && isRoomStillAvailable(selectPreAppointmentLocation))
  `

backend/controllers/appointments/selectCourtAppointmentRooms.js
```diff
@@ -286,6 +287,29 @@ const selectCourtAppointmentRoomsFactory = ({
         })
       }
 
+      const availableRooms = await availableSlotsService.getAvailableRooms(res.locals, {
```
JamieDraperUK: I could be wrong but not sure how well this is going to work. Ideally I think we need to do checks on each of the three time slots: 

- Pre start and end time
- Main start and end time
- Post start and end time

JamieDraperUK: It might be a little overkill to use the availableSlotService to do the three checks. Prob best to use the appointmentService and replicate the check we do in addCourtAppointmentService.js

jsrobertson: looks like using `existingEventsService.getAvailableLocationsForVLB` might be better actually.

backend/controllers/appointments/selectCourtAppointmentRooms.js
jsrobertson: Not quite the same unfortunately.  Have neatened it up though.

backend/controllers/appointments/selectCourtAppointmentRooms.js
JamieDraperUK: Ah, yes - that's the one I meant actually. Basically just replicate the initial check in addCourtAppointmentService.js

jsrobertson: Updated to use `getAvailableLocationsForVLB` from `existingEventsService`.

PR #832, opened 2020-03-11
NN-2376 prison contact config

JamieDraperUK:


backend/config.js
```diff
@@ -54,5 +54,11 @@ module.exports = {
     confirmBookingPrisonTemplateId: '391bb0e0-89b3-4aef-b11e-c6550b71fee8',
     confirmBookingCourtTemplateId: '7f44cd94-4a74-4b9d-aff8-386fec34bd2e',
     requestBookingCourtTemplateId: 'c1008f55-c228-4cad-b6fd-fe931c993855',
+    emails: {
+      WWI: {
+        omu: 'OMU.wandsworth@justice.gov.uk',
```
JamieDraperUK: I wonder if these should be secrets?

Nimphal: There are various contacts for prisons that are public, so I don't think I feel too bad about this. That being said, we will have to look at moving this away from the config in a future iteration anyway - when we add the next prison probably.

JamieDraperUK: I'm gonna move this into secrets.

PR #850, opened 2020-03-16
NN-2489: Start using create user id rather than audit user id

Nimphal:


Nimphal:
This has a safety check in that if the staff can't be found it shows '--', so the page won't break, it will temporarily not show the information. I had started on the road to doing what you suggest, but decided against as the impact is minimal.

PR #851, opened 2020-03-17
NN-2376 fix validation message

JamieDraperUK:
Provide validation messages for when only part of the time has been entered.

### DIFF
--- a/backend/controllers/appointments/requestBooking.js
+++ b/backend/controllers/appointments/requestBooking.js
@@ -7,23 +7,33 @@ const {
   app: { videoLinkEnabledFor },
 } = require('../../config')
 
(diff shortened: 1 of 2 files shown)


backend/controllers/appointments/requestBooking.js
```diff
@@ -7,23 +7,33 @@ const {
   app: { videoLinkEnabledFor },
 } = require('../../config')
 
-const validateStartEndTime = (date, startTime, endTime, errors) => {
+const isValidNumber = number => Number.isSafeInteger(Number.parseInt(number, 10))
```
JamieDraperUK: Will pull this into utils when I revert the buildTimeDate fix.

PR #854, opened 2020-03-17
NN-2451 Update error messaging for add VLB service

jsrobertson:


backend/controllers/appointments/addCourtAppointment.js
```diff
@@ -8,23 +8,36 @@ const { serviceUnavailableMessage } = require('../../common-messages')
 
 const addCourtAppointmentsFactory = (elite2Api, logError) => {
   const getValidationMessages = fields => {
```
JamieDraperUK: I think this now duplicates what I did for the previous PR, the one with the isNumber function. It it worth reusing anything?

PR #857, opened 2020-03-18
NN-2482 Add court layout template with court specific footer

jsrobertson:
New footer for court service pages with contact link which directs to auth contact page.
image


backend/controllers/appointments/confirmAppointment.js
```diff
@@ -14,6 +14,7 @@ const confirmAppointmentFactory = ({ elite2Api, appointmentsService, logError })
   const index = async (req, res) => {
```
jsrobertson: Used `authSource === 'nomis'` both in here and `videolinkBookingConfirmHearing`.  Pretty sure it's the only page where we re-use between standard add appointment and add court appointment now.

May be an idea to split these out so they are 2 separate things for when the court stuff gets moved to it's own service.  Not one for now though I don't think.

PR #870, opened 2020-03-26
GDPR-78 Adding retention reasons page

brightonsbox:
Currently this has no functionality, it is just
styling and setting up some basic tests.

### DIFF
--- a/prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/specs/RetentionReasonsSpecification.groovy
+++ b/prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/specs/RetentionReasonsSpecification.groovy
@@ -1,0 +1,53 @@
+package uk.gov.justice.digital.hmpps.prisonstaffhub.specs
+
+import org.junit.Rule
+import uk.gov.justice.digital.hmpps.prisonstaffhub.mockapis.Elite2Api
+import uk.gov.justice.digital.hmpps.prisonstaffhub.mockapis.OauthApi
+import uk.gov.justice.digital.hmpps.prisonstaffhub.mockapis.WhereaboutsApi
+import uk.gov.justice.digital.hmpps.prisonstaffhub.mockapis.mockResponses.ActivityResponse
+import uk.gov.justice.digital.hmpps.prisonstaffhub.mockapis.mockResponses.VisitsResponse
+import uk.gov.justice.digital.hmpps.prisonstaffhub.model.Caseload
+import uk.gov.justice.digital.hmpps.prisonstaffhub.model.TestFixture
+import uk.gov.justice.digital.hmpps.prisonstaffhub.model.UserAccount
+import uk.gov.justice.digital.hmpps.prisonstaffhub.pages.*
+
+import java.time.LocalDate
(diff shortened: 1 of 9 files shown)


prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/specs/RetentionReasonsSpecification.groovy
```diff
@@ -0,0 +1,53 @@
...
+    WhereaboutsApi whereaboutsApi = new WhereaboutsApi()
+
+    TestFixture fixture = new TestFixture(browser, elite2api, oauthApi, whereaboutsApi)
+
+    def "should load the retention reasons page"() {
+        setupTests()
+
+        given: "I am on the retention reasons page"
+        to RetentionReasonsPage
+    }
```
brightonsbox: Currently is just testing that the page loads, no functionality to test yet

sass/components/_offender-basic-details.scss
```diff
@@ -0,0 +1,38 @@
+.bold {
+  font-weight: bold
+}
+
+.basic-details {
+  border-top: 1px solid #b1b4b6;
+  border-bottom: 1px solid #b1b4b6;
+  padding-top: 15px;
+  padding-bottom: 15px;
```
andrewrlee: you shouldn't really need to add so much custom styling if using gds - there are custom built classes that you can use. https://design-system.service.gov.uk/styles/spacing/

andrewrlee: similar with typography too

views/retentionReasons.njk
```diff
@@ -0,0 +1,158 @@
...
+        fieldset: {
+          legend: {
+            text: subtitle,
+            classes: "govuk-fieldset__legend--m"
+          }
+        },
+        hint: {
+          text: instructions 
+        },
+        items: [
```
andrewrlee: should these be pulled from the data compliance service?

andrewrlee: just thinking that if you wanted to change these that would only be a release to data compliance not here

brightonsbox: 👍 yes that's the end goal, but just wanted to start with something that visually matches what I need

sass/components/_offender-basic-details.scss
JamieDraperUK: Be worth looking at the other pages Jon.

brightonsbox: This is matching Pathfinder, so for consistency wanted to keep it as close to that as possible.  I'll take a look to reduce this though.

views/retentionReasons.njk
```diff
@@ -0,0 +1,158 @@
...
+                  src="/app/images/{{ offenderBasics.offenderNo }}/data"
+                  alt="Headshot of {{ offenderBasics.firstName }} {{ offenderBasics.lastName }}"
+                  style="float:left"
+                  height="110"
+                  width="90"
+                  data-qa="offender-image"
+          />
+        </div>
+  
+        <div>
```
JamieDraperUK: In order to use the correct font here you'd need to use <div class="govuk-body">govuk-body</div>

PR #871, opened 2020-03-26
GDPR-78 Adding Data Compliance API

brightonsbox:


backend/api/dataComplianceApi.test.js
```diff
@@ -0,0 +1,43 @@
...
+const dataComplianceApi = dataComplianceApiFactory(client)
+const context = { some: 'context' }
+
+describe('data compliance api', () => {
+  describe('get offender retention reasons', () => {
+    const response = { some: 'response' }
+    let actual
+
+    beforeEach(() => {
+      client.get = jest.fn().mockReturnValue({ then: () => response })
```
andrewrlee: `mockResolvedValue`

brightonsbox: Would that work here? This is returning an object that has a method `then` so I don't think `mockResolvedValue(response)` would work.  Is there some other way it can be used?

andrewrlee: yes - mockResolvedValue wraps the passed arg in a promise - which has `then` and is what you are trying to emulate here

PR #872, opened 2020-03-26
NN-2523: Tighten buildDateTime function, make sure appointments without end time still work

Nimphal:


src/dateHelpers.js
```diff
@@ -10,9 +10,12 @@ const DayMonthYear = dateTime => moment(dateTime, DATE_TIME_FORMAT_SPEC).format(
...
 const buildDateTime = ({ date, hours, minutes }) => {
-  const time = date && hours && minutes && moment(date, DAY_MONTH_YEAR)
-
-  return time && time.hour(Number(hours)).minutes(Number(minutes))
+  const time =
+    date &&
+    Number.isSafeInteger(Number.parseInt(hours, 10)) &&
+    Number.isSafeInteger(Number.parseInt(minutes, 10)) &&
+    moment(date, DAY_MONTH_YEAR)
+  return time ? time.hour(Number(hours)).minutes(Number(minutes)) : ''
```
Nimphal: In the previous attempt to introduce this safety, this function was returning `false` if passed empty strings, which was then passed onto the API, which in turn errorred out as it was not expecting a boolean for the endTime parameter.

PR #874, opened 2020-03-27
NN-2524: Add prisoner not listed request booking info page

Nimphal:
Page is not linked-to from anywhere yet, part of a larger journey.


backend/routes/appointments/requestBookingRouter.js
```diff
@@ -27,6 +27,13 @@ const controller = ({ logError, notifyClient, whereaboutsApi, oauthApi, elite2Ap
   router.get('/enter-offender-details', enterOffenderDetails)
   router.post('/create-booking-request', createBookingRequest)
   router.get('/confirmation', confirm)
+  router.get('/prisoner-not-listed', async (req, res) => {
```
brightonsbox: Should this not go in `requestBooking.js`?

Nimphal: That's the controller, and we usually put logic around the request and data crunching in there. Since this is purely serving a template, we keep it in the router as there's no need for a special controller for it.

PR #878, opened 2020-03-30
GDPR-78 Presenting existing reasons and posting updates

brightonsbox:


backend/api/dataComplianceApi.test.js
```diff
@@ -6,38 +6,75 @@ const context = { some: 'context' }
...
+      actual = await dataComplianceApi.getOffenderRetentionRecord(context, 'A1234BC')
     })
 
     it('should call the correct endpoint', () => {
       expect(client.get).toBeCalledWith(context, '/retention/offenders/A1234BC')
     })
 
-    it('should return response from endpoint', () => {
-      expect(actual).toEqual(response)
+    it('handles jetty bug where gzip encoding causes etag suffix', async () => {
```
brightonsbox: Was encountering this with the Wiremock tests so have handled it to be safe https://github.com/eclipse/jetty.project/issues/1416

package.json
```diff
@@ -79,6 +79,7 @@
     "passport-oauth2": "^1.4.0",
     "path-to-regexp": "^2.4.0",
     "prop-types": "^15.7.2",
+    "qs": "^6.5.2",
```
brightonsbox: Adding explicit dependency here (although it's already transitively used).

We currently use `bodyParser.urlencoded({ extended: **false** })` so posted forms are parsed with the query-string library rather than qs.  This means that tricks like form input arrays to group inputs doesn't work out of the box.  I've opted for using qs directly where I've needed it since changing to `extended: true` across the board seems quite a dangerous change.

backend/controllers/retentionReasons.js
```diff
@@ -42,7 +72,23 @@ const retentionReasonsFactory = (elite2Api, dataComplianceApi, logError) => {
 
   const index = async (req, res) => renderTemplate(req, res)
 
-  return { index }
+  const post = async (req, res) => {
+    try {
+      const { offenderNo } = req.params
+      const { reasons, version } = qs.parse(req.body)
```
brightonsbox: Using qs here to very simply group reason codes and details together into an array.  Without this, I'd need some custom JS to parse the form inputs in the request body, handling things like singular checkboxes generating strings, not arrays of strings...

backend/api/axios-config-decorators.js
```diff
@@ -2,15 +2,15 @@ const contextProperties = require('../contextProperties')
...
+  const customHeaders = contextProperties.getCustomRequestHeaders(context)
   const accessToken = contextProperties.getAccessToken(context)
-  return accessToken
-    ? { authorization: `Bearer ${accessToken}`, ...paginationHeadersWithResultsLimit }
-    : paginationHeadersWithResultsLimit
+
+  return {
+    ...customHeaders,
+    ...paginationHeaders,
+    ...(resultsLimit && { 'page-limit': resultsLimit.toString() }),
```
petergphillips: didn't realise that deconstructing `false` would then be ignored - code lot simpler :+1:

backend/api/dataComplianceApi.js
```diff
@@ -1,20 +1,38 @@
...
 
 const dataComplianceApiFactory = client => {
-  const processResponse = context => response => {
-    contextProperties.setResponsePagination(context, response.headers)
-    return response.body
+  const getResponseBody = response => response.body
+
+  const includeETagInBody = response => ({
+    ...response.body,
+    etag: response.headers.etag.replace('--gzip', ''),
```
petergphillips: what's going on here?

petergphillips: ah found the explanation lower down

PR #882, opened 2020-03-31
GDPR-78 A few small updates e.g. to display last update details

brightonsbox:


backend/api/dataComplianceApi.js
```diff
@@ -5,7 +5,7 @@ const dataComplianceApiFactory = client => {
 
   const includeETagInBody = response => ({
     ...response.body,
-    etag: response.headers.etag.replace('--gzip', ''),
+    etag: response.headers.etag.replace(/--gzip|W\//gi, ''),
```
brightonsbox: To handle weak etags

backend/controllers/retentionReasons.js
```diff
@@ -1,24 +1,27 @@
+/* eslint-disable consistent-return */
```
brightonsbox: A few helper methods here can return null

backend/controllers/retentionReasons.js
```diff
@@ -1,24 +1,27 @@
...
   const sortReasonsByDisplayOrder = reasons => reasons.sort((r1, r2) => (r1.displayOrder > r2.displayOrder ? 1 : -1))
 
-  const getAgencyDescription = (agencies, agencyId) => {
-    const agency = agencies.find(a => a.agencyId === agencyId)
-    return agency ? agency.description : 'Unknown'
-  }
-
   const getOffenderUrl = offenderNo => `${dpsUrl}offenders/${offenderNo}`
 
+  const formatDate = date => date && formatTimestampToDate(date)
```
andrewrlee: these format methods handle null anyway?

backend/controllers/retentionReasons.js
andrewrlee: better to not blanket ignore the whole file as the warnings. These are generally easy to fix, which places in particular?

backend/controllers/retentionReasons.js
andrewrlee: e.g: https://github.com/ministryofjustice/prisonstaffhub/pull/882/files#diff-d2bd71a8ef7316942b1e5a4166909718L140

brightonsbox: Ah right you are.  For some reason I'd convinced myself that they wouldn't handle it!

backend/controllers/retentionReasons.js
brightonsbox: I'll have another look, the linter wasn't having various different arrangements of one function.

PR #883, opened 2020-03-31
GDPR-78 Addressing comments

brightonsbox:


backend/controllers/retentionReasons.js
```diff
@@ -83,7 +80,7 @@ const retentionReasonsFactory = (elite2Api, dataComplianceApi, logError) => {
         retentionReasons: reasons.filter(reason => reason.reasonCode),
       }
 
-      dataComplianceApi
+      return await dataComplianceApi
```
andrewrlee: Bit confusing mixing `await` and promises..

can get rid of the `then` and `catch` if using await. (any exception would be caught by the `} catch (error) {`)
```js
    await dataComplianceApi.putOffenderRetentionRecord(res.locals, offenderNo, optionsSelected, version)
    return res.redirect(getOffenderUrl(offenderNo))
       
  } catch (error) {
      return renderError(req, res, error)
    }
```

backend/controllers/retentionReasons.js
```diff
@@ -11,18 +10,16 @@ const retentionReasonsFactory = (elite2Api, dataComplianceApi, logError) => {
...
+    existingRecord
+      ? {
+          version: existingRecord.etag,
+          timestamp: formatTimestampToDateTime(existingRecord.modifiedDateTime),
+          user: existingRecord.userId,
+        }
+      : {}
 
   const flagReasonsAlreadySelected = (retentionReasons, existingRecord) => {
     const existingReasons = existingRecord && existingRecord.retentionReasons
```
andrewrlee: Rather than checking twice for existence here and in the `matchedReason` bit below - this could always return an empty list

backend/controllers/retentionReasons.js
```diff
@@ -11,25 +10,23 @@ const retentionReasonsFactory = (elite2Api, dataComplianceApi, logError) => {
...
-      timestamp: formatDateTime(existingRecord.modifiedDateTime),
-      user: existingRecord.userId,
-    }
+    existingRecord
+      ? {
+          version: existingRecord.etag,
+          timestamp: formatTimestampToDateTime(existingRecord.modifiedDateTime),
+          user: existingRecord.userId,
+        }
+      : {}
```
andrewrlee: an option here would have been to use destructuring: 
```js
 (existingRecord = {}) => ({ ...existingRecord, timestamp: formatDateTime(existingRecord.modifiedDateTime) })
```
But its kind of nice seeing the structure of the object anyway..

backend/controllers/retentionReasons.js
```diff
@@ -11,25 +10,23 @@ const retentionReasonsFactory = (elite2Api, dataComplianceApi, logError) => {
...
 
   const flagReasonsAlreadySelected = (retentionReasons, existingRecord) => {
-    const existingReasons = existingRecord && existingRecord.retentionReasons
+    const existingReasons = existingRecord.retentionReasons || []
```
andrewrlee: this isn't entirely the same as previously `existingRecord` could have been undefined.. 

could provide a default param: `(retentionReasons, existingRecord = {}) => {`
or maybe pass through only the existing reasons and default them..

backend/controllers/retentionReasons.js
brightonsbox: Nice!

PR #889, opened 2020-04-01
NN-2526: Include suspended count to attendance statistics dashboard

Nimphal:
Screenshot from 2020-04-01 14-55-39


Nimphal:
Do you mean for the lists of offenders? That's a separate story.

PR #897, opened 2020-04-09
NN-2552 Add prisoner profile header links for Incentive Level and Categorisation

jsrobertson:
image

### DIFF
--- a/backend/services/prisonerProfileService.js
+++ b/backend/services/prisonerProfileService.js
@@ -1,8 +1,13 @@
 const moment = require('moment')
 const { properCaseName } = require('../utils')
 const alertFlagValues = require('../shared/alertFlagValues')
+const {
+  apis: {
+    categorisation: { ui_url: categorisationUrl },
+  },
+} = require('../config')
 
-module.exports = (elite2Api, keyworkerApi) => {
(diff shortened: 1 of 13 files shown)


backend/services/prisonerProfileService.js
```diff
@@ -16,21 +21,39 @@ module.exports = (elite2Api, keyworkerApi) => {
...
 
-    const [iepDetails, keyworkerSessions, keyworkerDetails] = await Promise.all([
-      elite2Api.getIepSummary(context, [bookingId]),
-      elite2Api.getCaseNoteSummaryByTypes(context, { type: 'KA', subType: 'KS', numMonths: 1, bookingId }),
-      keyworkerApi.getKeyworkerByCaseloadAndOffenderNo(context, agencyId, offenderNo),
-    ])
+    const canViewInactivePrisoner = userRoles && userRoles.some(role => role.roleCode === 'INACTIVE_BOOKINGS')
+    const offenderInCaseload = userCaseloads.some(caseload => caseload.caseLoadId === agencyId)
+
+    const isCatToolUser = Boolean(
```
jsrobertson, line 40: Suspect we may need to know this elsewhere at some point but cross that bridge when we come to it, I think.

PR #904, opened 2020-04-16
NN-2457 Add case note, adjudications and visits data to quick look

jsrobertson:
image

### DIFF
--- a/backend/controllers/prisonerProfile/prisonerQuickLook.js
+++ b/backend/controllers/prisonerProfile/prisonerQuickLook.js
@@ -4,6 +4,7 @@ const {
   app: { notmEndpointUrl: dpsUrl },
 } = require('../../config')
 const { formatTimestampToDate, formatCurrency } = require('../../utils')
+const formatAward = require('../../shared/formatAward')
 
 module.exports = ({ prisonerProfileService, elite2Api, logError }) => async (req, res) => {
   try {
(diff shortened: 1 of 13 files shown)


backend/shared/formatAward.js
```diff
@@ -0,0 +1,37 @@
+const { formatTimestampToDate, formatCurrency } = require('../utils')
```
jsrobertson, line 1: Existing file from new-nomis-ui

backend/controllers/prisonerProfile/prisonerQuickLook.js
```diff
@@ -3,53 +3,105 @@ const { serviceUnavailableMessage } = require('../../common-messages')
...
+      ],
+      balanceDetails: [
+        { label: 'Spends', value: formatCurrency(balanceData.spends, balanceData.currency) },
+        { label: 'Private', value: formatCurrency(balanceData.cash, balanceData.currency) },
+        { label: 'Savings', value: formatCurrency(balanceData.savings, balanceData.currency) },
+      ],
+      caseNoteAdjudications: {
+        details: [
+          { label: 'Incentive level warnings', value: negativeCaseNotes.count },
+          { label: 'Incentive Encouragements', value: positiveCaseNotes.count },
```
petergphillips, line 67: do you need a capital `E` here?

backend/utils.js
```diff
@@ -199,6 +199,20 @@ const hyphenatedStringToCamel = string =>
 const formatCurrency = (number, currency = 'GBP') =>
   number ? number.toLocaleString('en-GB', { style: 'currency', currency }) : ''
 
+const capitalizeUppercaseString = string =>
+  string
+    ? string
+        .split(' ')
+        .map(name => capitalize(name))
```
petergphillips, line 206: we normally just capitalise the first word rather than all of them?

backend/controllers/prisonerProfile/prisonerQuickLook.js
jsrobertson: Unbelievably, yes 😢

backend/utils.js
jsrobertson: Could probably name this better, meant for names rather than anything else.  Will change in next PR as I've still got some stuff to do.

PR #905, opened 2020-04-17
NN-2457 Add todays scheduled events to prisoner profile quick look

jsrobertson:
image

image


backend/shared/filterActivitiesByPeriod.js
```diff
@@ -0,0 +1,40 @@
+const moment = require('moment')
```
jsrobertson, line 1: Copied over from new-nomis-ui.

PR #920, opened 2020-04-24
Nn 2528 suspended list

Nimphal:
Add link to the dashboard stats through to a new page that lists the suspended offenders for the period(s) and date range and their attendance outcome.

### DIFF
--- a/backend/controllers/attendance/attendanceStatistics.js
+++ b/backend/controllers/attendance/attendanceStatistics.js
@@ -11,6 +11,7 @@ const {
   pascalToString,
   readableDateFormat,
   stripAgencyPrefix,
+  properCaseName,
 } = require('../../utils')
 
 const attendanceReasonStatsUrl = '/manage-prisoner-whereabouts/attendance-reason-statistics'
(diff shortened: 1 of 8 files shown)


backend/controllers/attendance/attendanceStatistics.js
```diff
@@ -326,9 +327,140 @@ const attendanceStatisticsFactory = (oauthApi, elite2Api, whereaboutsApi, logErr
...
+          periodsList.map(async periodItem => {
+            await Promise.all(
+              Object.keys(data).map(async date => {
+                const attendances = await whereaboutsApi.getAttendanceForBookings(res.locals, {
+                  agencyId,
+                  period: periodItem,
+                  bookings: data[date],
+                  date,
+                })
+                suspendedAttendancesList.push(...attendances.attendances)
```
Nimphal, line 387: This is a bit involved in order to accommodate the APIs current interface, which takes only one date and one period. Potential work to refactor the API endpoint to accommodate this use case will be done outside of this ticket.

Nimphal: The maximum amount of calls to the whereabouts API from this page is 28 (14 days, 2 periods). This is not ideal, but is a rare use case where every single day would have suspended schedules in it.

backend/controllers/attendance/attendanceStatistics.js
```diff
@@ -326,9 +327,140 @@ const attendanceStatisticsFactory = (oauthApi, elite2Api, whereaboutsApi, logErr
...
+        toDate: formattedToDate,
+        period: period === 'AM_PM' ? '' : period,
+      })
+
+      const suspendedActivites = scheduledActivities.filter(activity => activity.suspended)
+      const totalOffenders = new Set(suspendedActivites.map(activity => activity.bookingId)).size
+
+      // Create mapping between date and a set of bookings with suspended activities on that date
+      // This mapping will be used to get the relevant attendances for each date
+      const suspendedBookingsPerEventDate = {}
```
JamieDraperUK, line 366: Might be more consistent to use reduce and the spread operator here.

Nimphal: I was kind of hoping to leave it a bit less... clever? Whatever the word is, just because it's kind of involved and like this tells the story a bit more clearly. But happy to refactor if you think the opposite :D

backend/controllers/attendance/attendanceStatistics.js
```diff
@@ -326,9 +327,140 @@ const attendanceStatisticsFactory = (oauthApi, elite2Api, whereaboutsApi, logErr
...
+      suspendedActivites.forEach(activity => {
+        const date = moment(activity.startTime).format('YYYY-MM-DD')
+        if (!suspendedBookingsPerEventDate[date]) {
+          suspendedBookingsPerEventDate[date] = new Set()
+        }
+        suspendedBookingsPerEventDate[date].add(activity.bookingId)
+      })
+
+      const getAttendances = async (periodsList, data) => {
```
JamieDraperUK, line 375: My eyes 😢

Nimphal: I know :/

PR #947, opened 2020-05-13
NN-2579 Add border to top of summary lists when there are multiple in a section

jsrobertson:
...except first one

Also capitalised place of birth.

**Before:** 
image

**After:** (see above Sexual orientation)
image


backend/controllers/prisonerProfile/prisonerQuickLook.js
```diff
@@ -57,6 +57,8 @@ module.exports = ({ prisonerProfileService, elite2Api, logError }) => async (req
   const prisoner = prisonerData && prisonerData[0]
   const { morningActivities, afternoonActivities, eveningActivities } = filterActivitiesByPeriod(todaysEvents)
 
+  console.log({ prisoner })
```
petergphillips, line 60: :scream:

jsrobertson: 😱

jsrobertson:
> Not sure how eslint allowed `console.log` in there?

Not sure either, will have a look into it.  Removed it though.

PR #950, opened 2020-05-14
Nn 2547 move prisoner alerts

Nimphal:
Transfers alerts tab from new nomis ui. Adds implementation for pagination as a separate service.


views/prisonerProfile/partials/prisonerProfileTabs.njk
```diff
@@ -14,7 +14,7 @@
   <ul class="govuk-tabs__list">
     {% for tab in tabs %}
       {% set fullTabPath = '/prisoner/' + prisonerProfileData.offenderNo + tab.path %}
-      {% set isSelected = ' govuk-tabs__list-item--selected' if currentUrlPath === fullTabPath %}
+      {% set isSelected = ' govuk-tabs__list-item--selected' if currentUrlPath.split('?')[0] === fullTabPath %}
```
Nimphal, line 17: This is to tackle urls with query params in them.

backend/api/elite2Api.js
```diff
@@ -93,6 +93,11 @@ const elite2ApiFactory = client => {
   const getAlerts = (context, { agencyId, offenderNumbers }) =>
     post(context, `/api/bookings/offenderNo/${agencyId}/alerts`, offenderNumbers)
 
+  const getAlertsForBooking = (context, { bookingId, query }, headers) => {
+    contextProperties.setCustomRequestHeaders(context, headers)
+    return get(context, `/api/bookings/${bookingId}/alerts${query}`)
+  }
+
```
brightonsbox, line 100: I just noticed that we don't have any tests for elite2Api.js, but most of the other API clients have them. So nothing to do with your PR really, I just wondered why?

Nimphal: The ones that have tests test logic which manipulates the response or the request in a way. Elite2 and Whereabouts purely pass on the response to the controller, whose tests mock it out.

Nimphal: That isn't to say we couldn't add some tests, I'm sure there are things to test :D

PR #952, opened 2020-05-14
Nn 2547 fix total alerts count

Nimphal:


backend/controllers/prisonerProfile/prisonerAlerts.js
```diff
@@ -40,14 +40,13 @@ module.exports = ({ prisonerProfileService, referenceCodesService, paginationSer
...
+    const [prisonerProfileData, alertTypes] = await Promise.all([
       prisonerProfileService.getPrisonerProfileData(res.locals, offenderNo),
       referenceCodesService.getAlertTypes(res.locals),
-      elite2Api.getAlertsForBooking(res.locals, { bookingId, query }, headers),
     ])
 
-    const totalAlerts =
-      active === 'INACTIVE' ? prisonerProfileData.inactiveAlertCount : prisonerProfileData.activeAlertCount
+    const alerts = await elite2Api.getAlertsForBooking(res.locals, { bookingId, query }, headers)
+    const totalAlerts = res.locals.responseHeaders['total-records']
```
Nimphal, line 49: The previous calculation did not take into account additional filters, so would have displayed 'Showing 1 of 20 from 20' even when there were 6 results in the page if the offender had 20 active/inactive alerts

PR #953, opened 2020-05-14
DT-840: Add token verification health check and tidy

petergphillips:


.circleci/config.yml
```diff
@@ -214,7 +214,7 @@ jobs:
       - restore_cache:
           key: dependency-cache-{{ checksum "package.json" }}
       - run:
-          name: install-yarn
+          name: install-npm
```
petergphillips, line 217: left over from previous PR

backend/services/healthCheck.js
```diff
@@ -1,7 +1,8 @@
 const { serviceCheckFactory } = require('../controllers/healthCheck')
 
 const service = (name, url) => {
-  const check = serviceCheckFactory(name, url)
+  const healthUrl = `${url.replace(/\/$/, '')}/health/ping`
```
petergphillips, line 4: switch to health/ping and fix trailing slashes

.circleci/config.yml
```diff
@@ -101,7 +101,7 @@ jobs:
           name: Run Tests
           command: |
              set -e
-             npm test --maxWorkers=2
+             npm test -- --maxWorkers=2
```
petergphillips, line 104: so `npm` needs an extra `--` to pass the parameter through

feature.env
```diff
@@ -0,0 +1,12 @@
+PORT=3006
+OAUTH_ENDPOINT_URL=http://localhost:19090/auth/
```
petergphillips, line 2: feature config in package.json was getting too big, so moved to separate env file

package.json
```diff
@@ -163,7 +163,7 @@
   },
   "scripts": {
     "start": "node backend/index.js",
-    "start-feature": "PORT=3006 OAUTH_ENDPOINT_URL=http://localhost:19090/auth/ API_ENDPOINT_URL=http://localhost:18080/ NN_ENDPOINT_URL=http://localhost:20200/ API_DATA_COMPLIANCE_ENDPOINT_URL=http://localhost:18084 API_COMMUNITY_ENDPOINT_URL=http://localhost:18083 API_WHEREABOUTS_ENDPOINT_URL=http://localhost:18082/ VIDEO_LINK_ENABLED_FOR=WWI WANDSWORTH_OMU_EMAIL=test@local WANDSWORTH_VLB_EMAIL=test@local node backend/index.js",
+    "start-feature": "export $(cat feature.env) && node backend/index.js",
```
petergphillips, line 166: moved variables to separate file

html-template/index.html
```diff
@@ -4,11 +4,6 @@
   <meta charset="utf-8">
   <meta name="theme-color" content="#000000">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <!--
```
petergphillips, line 7: manifest was causing 401s to auth all the time, so removed as not needed (only used for android home screen apparently anyway and out of the box react that should be removed)

PR #954, opened 2020-05-14
NN-2579 Updated listener suitable and recognised display conditions

jsrobertson:


jsrobertson:
> LGTM

Thanks @Nimphal .  Could you re-approve please? I added another test.

PR #956, opened 2020-05-15
Nn 2614 setup cypress

Nimphal:


Dockerfile
```diff
@@ -20,15 +20,15 @@ RUN mkdir -p /app
 WORKDIR /app
 ADD . .
 
-RUN yarn --frozen-lockfile && \
```
petergphillips, line 23: think this should have been replaced with
```
npm ci --no-audit
```

Dockerfile
```diff
@@ -20,15 +20,15 @@ RUN mkdir -p /app
...
-    yarn record-build-info
+    npm run record-build-info
 
 ENV PORT=3000
 
 EXPOSE 3000
 RUN chown -R appuser:appgroup /app
 USER 2000
-CMD [ "yarn", "start" ]
+CMD [ "npm", "run", "start" ]
```
petergphillips, line 34: can just be ````
CMD [ "npm", "start" ]
```

PR #959, opened 2020-05-18
NN-2629: Pass agency id as part of url, refactor integration test into cypress

Nimphal:


integration-tests/integration/attendance/attendanceChangesPage.spec.js
```diff
@@ -0,0 +1,41 @@
+const AttendanceChangesPage = require('../../pages/attendance/attendanceChangesPage')
+
+context('A user can view attendance changes', () => {
+  beforeEach(() => {
```
JamieDraperUK, line 4: Looks much better. Where is the WireMock data stubbed for this test?

JamieDraperUK: Ignore that I see the setup below.

PR #969, opened 2020-05-21
DT-840: Switch auth order round as not needed for static content

petergphillips:


Dockerfile
```diff
@@ -20,6 +20,9 @@ RUN mkdir -p /app
 WORKDIR /app
 ADD . .
 
+# Remove cypress as it takes ~1 minute to download and install
+RUN sed -ie 's/"cypress[^ ].*//g' /app/package.json
```
petergphillips, line 24: copied from use of force

backend/index.js
```diff
@@ -30,16 +30,15 @@ app.set('view engine', 'njk')
...
+app.use(setupStaticContent({ app }))
 app.use(setupWebSession())
 app.use(setupAuth({ oauthApi: apis.oauthApi }))
-app.use(setupSass())
 app.use(setupWebpackForDev())
 
-app.use(setupRedirects())
 app.use(routes(apis))
 
-app.use(setupStaticContent({ app }))
```
petergphillips, line 41: move stuff that doesn't need a auth session above

docker-compose-test.yaml
```diff
@@ -2,7 +2,7 @@ version: '3.1'
 services:
   prisonstaffhub-wiremock:
     image: rodolpheche/wiremock
-    container_name: prisonstaffhub
+    container_name: prisonstaffhub-wiremock
```
petergphillips, line 5: rename so doesn't conflict with `prisonstaffhub`

html-template/index.html
```diff
@@ -6,42 +6,23 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
   <!--[if !(IE)]><!-->
-  <link rel='stylesheet' href='<%= htmlWebpackPlugin.options.publicUrl %>/styles/grid.css' type='text/css'>
```
petergphillips, line 9: never set so removed

package.json
```diff
@@ -50,12 +50,10 @@
     "contentful": "^7.3.0",
     "cookie-session": "^2.0.0-beta.3",
     "csv-parse": "^4.4.6",
-    "cypress-multi-reporters": "^1.4.0",
```
petergphillips, line 53: moved into dev dependencies

backend/index.js
```diff
@@ -31,18 +31,18 @@ app.set('view engine', 'njk')
 app.use(setupBodyParsers())
 app.use(setupHealthChecks())
 app.use(setupWebSecurity())
+app.use(setupRedirects())
+app.use(setupSass())
+app.use(setupStaticContent({ app }))
```
JamieDraperUK, line 36: This no longer takes **app** as a parameter now that the nunjuck setup has been pulled out.

PR #970, opened 2020-05-21
NN-2275 global template current user

JamieDraperUK:


backend/utils/nunjucksSetup.js
```diff
@@ -137,4 +137,6 @@ module.exports = (app, path) => {
   njkEnv.addGlobal('notmUrl', config.app.notmEndpointUrl)
   njkEnv.addGlobal('oauthUrl', config.apis.oauth2.url)
   njkEnv.addGlobal('googleAnalyticsId', config.analytics.googleAnalyticsId)
+
+  return njkEnv
```
Nimphal, line 141: This is feels like an oversight :D

JamieDraperUK: My thoughts too!

PR #973, opened 2020-05-22
Nn 2541 last appointment and cell location

Coupey:


backend/controllers/appointments/bulkAppointmentsUpload.js
```diff
@@ -39,6 +39,17 @@ const bulkAppointmentsUploadFactory = (csvParserService, offenderLoader, logErro
     return uploadedList.filter(x => matchingPrisonerNumbers.indexOf(x) < 0)
   }
 
+  const getLocations = async (context, prisonerList) => {
+    return Promise.all(
+      prisonerList.map(async prisoner => {
+        const prisonerToUpdate = prisoner
+        const location = await offenderLoader.offenderLocation(context, prisoner.assignedLivingUnitId)
+        prisonerToUpdate.cellNo = location.description
+        return prisonerToUpdate
```
Nimphal, line 48: I think this could be simplified with destructuring:
```
prisonerList.map(async prisoner => {
        const location = await offenderLoader.offenderLocation(context, prisoner.assignedLivingUnitId)
        return {
                     ...prisoner,
                     cellNo: location.description
         }
```

I've not run the above, mind you, but something like it.

backend/controllers/appointments/bulkAppointmentsUpload.js
```diff
@@ -63,11 +74,14 @@ const bulkAppointmentsUploadFactory = (csvParserService, offenderLoader, logErro
...
           )
 
+          prisonersDetails = await getLocations(res.locals, prisonersDetails)
+
           const prisonerList = prisonersDetails.map(prisoner => ({
             bookingId: prisoner.bookingId,
             offenderNo: prisoner.offenderNo,
             firstName: prisoner.firstName,
             lastName: prisoner.lastName,
+            cellNo: prisoner.cellNo,
```
Nimphal, line 84: You could do the location lookup as part of this, saving mapping over the list twice and modifying it:
```const prisonerList = prisonersDetails.map(prisoner => (
   const location = await elite2Api.getLocation(context, prisoner.assignedLivingUnitId)
   return {
        bookingId: prisoner.bookingId,
        bookingId: prisoner.bookingId,
        offenderNo: prisoner.offenderNo,
        offenderNo: prisoner.offenderNo,
        firstName: prisoner.firstName,
        firstName: prisoner.firstName,
        lastName: prisoner.lastName,
        lastName: prisoner.lastName,
        cellNo: location.description,
        }
        )
```

This will also eliminate the need to modify the loader, or have a separate call to get the locations.

PR #978, opened 2020-05-22
DT-840: Disable dev webpack when running int tests

petergphillips:


Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN mkdir -p /app
 WORKDIR /app
 ADD . .
 
-RUN npm ci --no-audit && \
+RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit && \
```
petergphillips, line 23: @andrewrlee better way of not installing cypress rather than excluding from package.json

.circleci/config.yml
```diff
@@ -89,7 +89,7 @@ jobs:
           key: node_modules-{{ checksum "package-lock.json" }}-{{ checksum "package.json" }}
       - run:
           name: Install Dependencies
-          command: npm ci --no-audit
+          command: CYPRESS_INSTALL_BINARY=0 npm ci --no-audit
```
petergphillips, line 92: don't need cypress at this step

PR #980, opened 2020-05-22
DT-840: static content no auth

petergphillips:


backend/setupStaticContent.js
```diff
@@ -4,7 +4,7 @@ const path = require('path')
 const router = express.Router()
 
 module.exports = () => {
-  router.use(express.static(path.join(__dirname, '../build'))) // << this needed?
+  router.use('/bundle.js', express.static(path.join(__dirname, '../build/bundle.js')))
```
petergphillips, line 7: only want bundle.js served here, we want to protect index.html otherwise we don't get the login prompt for react part of PSH

package.json
```diff
@@ -127,13 +125,15 @@
     "chai": "^4.1.2",
     "copy-webpack-plugin": "^5.0.0",
     "css-loader": "^1.0.0",
-    "cypress": "^4.5.0",
+    "cypress": "^4.6.0",
+    "cypress-multi-reporters": "^1.4.0",
```
petergphillips, line 129: moved to dev dependencies

backend/index.js
```diff
@@ -33,15 +33,14 @@ nunjucksSetup(app, path)
 app.use(setupBodyParsers())
 app.use(setupHealthChecks())
 app.use(setupWebSecurity())
+app.use(setupRedirects())
+app.use(setupSass())
+app.use(setupStaticContent())
```
petergphillips, line 38: all moved above the auth section

html-template/index.html
```diff
@@ -6,42 +6,23 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
   <!--[if !(IE)]><!-->
-  <link rel='stylesheet' href='<%= htmlWebpackPlugin.options.publicUrl %>/styles/grid.css' type='text/css'>
+  <link rel='stylesheet' href='/styles/grid.css' type='text/css'>
```
petergphillips, line 9: the `publicUrl` variable here isn't used in webpack

views/partials/layout.njk
```diff
@@ -48,7 +48,7 @@
 {# Run JavaScript at end of the
 <body>, to avoid blocking the initial render. #}
 <script src="/assets/govuk/all.js"></script>
-<script src="/static/js/main.js"></script>
```
petergphillips, line 51: must be some magic going on so that these paths work in dev - both https://whereabouts-dev.prison.service.justice.gov.uk/js/main.js and https://whereabouts-dev.prison.service.justice.gov.uk/static/js/main.js serve the same content.  However after my changes the `static` version requires the user to be logged in :shrug:

PR #1002, opened 2020-06-04
DT-840: Move login specification to cypress

petergphillips:


integration-tests/mockApis/auth.js
```diff
@@ -56,7 +56,7 @@ const logout = () =>
   stubFor({
     request: {
       method: 'GET',
-      urlPattern: '/auth/logout.*',
+      urlPath: '/auth/logout',
```
petergphillips, line 59: `urlPath` doesn't include any query parameters

PR #1008, opened 2020-06-08
Nn 2668 upgrade webpack 5

Nimphal:


webpack.config.js
```diff
@@ -18,15 +18,20 @@ const developmentEntries = isDevelopment
...
   new webpack.ContextReplacementPlugin(/moment[/\\]locale$/, /en/),
-  new CopyWebpackPlugin([
-    {
-      from: 'static',
-      to: 'static',
-    },
-  ]),
+  new webpack.ProvidePlugin({
+    process: 'process/browser',
+  }),
```
Nimphal, line 23: Vfile (which is a dependency of a dependency of a dependency) requires process to be shimmed in the browser for it to work.

PR #1010, opened 2020-06-08
DT-840: Fix token verification failure infinite loop and enable in dev

petergphillips:


backend/api/tokenVerificationApi.js
```diff
@@ -23,7 +23,7 @@ const tokenVerificationApiFactory = client => {
         log.debug(`Token verification result is ${result}`)
         return result
       })
-      .catch(processError)
+      .catch(() => false)
```
petergphillips, line 26: this was causing the promise not to resolve properly.  Fine not to throw error as has been logged

helm_deploy/values-dev.yaml
```diff
@@ -34,5 +34,5 @@ env:
   USE_OF_FORCE_URL: "https://dev.use-of-force.service.justice.gov.uk"
   USE_OF_FORCE_PRISONS: "WRI"
   TOKENVERIFICATION_API_URL: "https://token-verification-api-dev.prison.service.justice.gov.uk"
-  TOKENVERIFICATION_API_ENABLED: "false"
+  TOKENVERIFICATION_API_ENABLED: "true"
```
petergphillips, line 37: just enable in dev for now

PR #1025, opened 2020-06-12
NN-2641 sentence adjustments

JamieDraperUK:
Screenshot 2020-06-12 at 09 04 39


integration-tests/plugins/index.js
```diff
@@ -134,6 +134,8 @@ module.exports = on => {
     stubReleaseDatesOffenderNo: releaseDates => Promise.all([elite2api.stubPrisonerSentenceDetails(releaseDates)]),
     stubVerifyToken: (active = true) => tokenverification.stubVerifyToken(active),
     stubLoginPage: auth.redirect,
+    stubSentenceAdjustments: response => elite2api.stubGetSentenceAdjustments(response),
+    stubPrisonerDetail: response => elite2api.stubPrisonerDetail(response),
```
Nimphal, line 138: I think there's a just below which does stubOffenderBasicDetails. I'm probably wrong, but I think it stubs basic getDetails, which is what you're using for the bookingId

PR #1031, opened 2020-06-15
DCS-527 Hooking up covid unit pages to dashboard

andrewrlee:
The numbers will only become links if there is at least one case.

Also moving all list page urls to a prefix of `/current-covid-units/`
(The dashboard which is currently on `/current-covid-units` will still be accessible on the original url so
the new-nomis url doesn't need to be updated)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,7 +64,7 @@ executors:
     working_directory: ~/app
   builder:
     docker:
-      - image: circleci/node:12.16.1-buster-browsers
+      - image: circleci/node:12-buster-browsers
     working_directory: ~/app
 
 orbs:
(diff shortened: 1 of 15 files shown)


.circleci/config.yml
```diff
@@ -64,7 +64,7 @@ executors:
     working_directory: ~/app
   builder:
     docker:
-      - image: circleci/node:12.16.1-buster-browsers
+      - image: circleci/node:12-buster-browsers
```
andrewrlee, line 67: possibly controversial, but this doesn't tie the node version to a specific point release.
(so will actually upgrade to 12.18)

andrewrlee: (this was a bit of an accidental change but thought it might be worth leaving it in.)

PR #1039, opened 2020-06-15
DCS-539: warning icon made round using CSS

GurnankCheema:
Inserted CSS specifically for this element because there wasn't an option in the govuk warning component to target just the icon. Classes could be added to the component but only for the warning text not the icon.

### DIFF
--- a/sass/components/covid/dashboard.scss
+++ b/sass/components/covid/dashboard.scss
@@ -4,3 +4,9 @@
     width: 13%;
     float: left;
   }
+
+div[data-qa='refusingToShield'] .govuk-warning-text__icon {
+    padding-right: 10px;
+    padding-left: 10px;
+  }
+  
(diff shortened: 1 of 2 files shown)


sass/components/covid/dashboard.scss
```diff
@@ -3,4 +3,10 @@
     padding: 0 15px;
     width: 13%;
     float: left;
-  }
\ No newline at end of file
+  }
+
+div[data-qa='refusingToShield'] .govuk-warning-text__icon {
```
jsrobertson, line 8: Can see your predicament but it's best not to use anything other than styling classes for applying styles to an element.  If someone decides to change that `data-qa` tag in future, they're unlikely to know that it will have any effect on how something looks.

I'll help you have a look into what the actual problem is.

GurnankCheema: agreed

PR #1056, opened 2020-06-23
NN-2740 Fix new prisoner search alert checkbox combo bugs

jsrobertson:
- Fix bug where a checked **ACCT open** `HA` would check both **ACCT Open** `HA` and **ACCT post closure** `HA1` on refresh but only search for **ACCT open** `HA`.
- Fix checkbox selection and API search bugs with checkboxes which have multiple associated alert codes. e.g. **Risk to LGBT** has codes `RTP` and `RLG`.

### DIFF
--- a/backend/controllers/search/prisonerSearch.js
+++ b/backend/controllers/search/prisonerSearch.js
@@ -12,14 +12,14 @@ module.exports = ({ paginationService, elite2Api, logError }) => async (req, res
   const {
     location,
     keywords,
-    alerts: queryAlerts,
+    alerts,
     pageOffsetOption,
     view,
     sortFieldsWithOrder = 'lastName,firstName:ASC',
   } = req.query
 
-  const multipleAlertsSelected = Array.isArray(queryAlerts)
-  const alerts = multipleAlertsSelected ? queryAlerts.map(alert => alert.split(',')).flat() : queryAlerts
+  const multipleAlertsSelected = Array.isArray(alerts)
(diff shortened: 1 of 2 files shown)


backend/controllers/search/prisonerSearch.js
```diff
@@ -62,10 +65,12 @@ module.exports = ({ paginationService, elite2Api, logError }) => async (req, res
...
-      alertOptions: alertFlagValues.map(alertFlag => ({
-        value: alertFlag.alertCodes,
-        text: alertFlag.label,
-        checked: alertFlag.alertCodes.some(alert => alerts && alerts.includes(alert)),
+      alertOptions: alertFlagValues.map(({ alertCodes, label }) => ({
+        value: alertCodes,
+        text: label,
+        checked: multipleAlertsSelected
+          ? selectedAlerts.some(alert => alertCodes.includes(alert))
+          : Boolean(selectedAlerts) && alertCodes.includes(selectedAlerts.split(',')[0]),
```
Nimphal, line 73: Nice, took me a while to understand this. The split is because the alerts in the query could be `RTP,RLG` , so even though it's one checkbox (so not an array, not multipleAlertsSelected), it could be multiple alert codes. Right?

jsrobertson: Yeah it's not the nicest to be fair, felt like there was a better option but it wouldn't come to me.  Going to try what @andrewrlee suggested to remove some of the complexity.

PR #1063, opened 2020-06-29
Nn 2671 court booking cypress tests

Nimphal:


integration-tests/integration/appointments/addCourtAppointmentPrison.spec.js
```diff
@@ -0,0 +1,463 @@
...
+            .minutes(0)
+            .format('YYYY-MM-DDTHH:mm:ss'),
+          locationId: 2,
+        },
+      ],
+      location: 2,
+      date: tomorrow.format('yyyy-MM-DD'),
+    })
+
+    // cy.task('stubAppointmentsAtAgency', 'MDI', [
```
brightonsbox, line 422: Can this be deleted?

Nimphal: It can, good spot!

PR #1064, opened 2020-06-29
NN-2748 Prisoner profile access

jsrobertson:
If a prisoner is not in a users caseload and/or a user cannot view an active booking then:
- Hide Case notes tab
- Hide View adjudication history link
- Hide probation docs link
- Hide Create/Edit/Close alert buttons

### DIFF
--- a/views/prisonerProfile/partials/prisonerProfileTabs.njk
+++ b/views/prisonerProfile/partials/prisonerProfileTabs.njk
@@ -1,9 +1,9 @@
 {% set tabs = [
-    { text: 'Quick look', path: '' },
-    { text: 'Personal', path: '/personal' },
-    { text: 'Alerts', path: '/alerts' },
-    { text: 'Case notes', path: '/case-notes' },
-    { text: 'Sentence and release', path: '/sentence-and-release' }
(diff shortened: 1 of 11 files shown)


views/prisonerProfile/partials/prisonerProfileTabs.njk
```diff
@@ -13,13 +13,15 @@
...
   <ul class="govuk-tabs__list">
     {% for tab in tabs %}
-      {% set fullTabPath = '/prisoner/' + prisonerProfileData.offenderNo + tab.path %}
-      {% set isSelected = ' govuk-tabs__list-item--selected' if currentUrlPath.split('?')[0] === fullTabPath %}
-      <li class="{{'govuk-tabs__list-item' + isSelected }}">
-        <a class="govuk-tabs__tab" href="{{ fullTabPath }}">
-        {{ tab.text }}
-        </a>
-      </li>
+      {% if tab.visible %}
```
Nimphal, line 16: Is that a sneaky fix for the quicklook tab quirk?

jsrobertson: No but thanks for reminding me, will have a look at that now.

PR #1070, opened 2020-07-01
NN-2757 Updated criteria for what active contacts get shown

jsrobertson:
We only care about active addresses for a contact.  If there is an active address flagged as primary, show that, if not, show the most recently added active address.

### DIFF
--- a/backend/tests/prisonerPersonal.test.js
+++ b/backend/tests/prisonerPersonal.test.js
@@ -1066,11 +1066,7 @@ describe('prisoner personal', () => {
               phones: [{ number: '02222222222', type: 'MOB' }, { number: '033333333333', type: 'MOB', ext: '777' }],
             })
             .mockResolvedValueOnce({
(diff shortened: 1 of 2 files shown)


backend/tests/prisonerPersonal.test.js
```diff
@@ -1031,8 +1067,9 @@ describe('prisoner personal', () => {
             })
             .mockResolvedValueOnce({
               addresses: [
-                { ...primaryAddress, endDate: '2020-01-01', addressType: 'Business' },
-                { ...nonPrimaryAddress, addressType: 'Business' },
+                // { ...businessPrimary, endDate: '2020-01-01', addressType: 'Business' },
```
Nimphal, line 1070: Did you mean to leave this in?

jsrobertson: Nope :D. Will take it out

PR #1078, opened 2020-07-03
NN-2770 request offences using client creds

JamieDraperUK:


backend/controllers/prisonerProfile/prisonerSentenceAndRelease.js
```diff
@@ -4,15 +4,17 @@ const courtCasesViewModel = require('./sentenceAndReleaseViewModels/courtCasesVi
 const { serviceUnavailableMessage } = require('../../common-messages')
 const { readableDateFormat } = require('../../utils')
 
-module.exports = ({ prisonerProfileService, elite2Api, logError }) => async (req, res) => {
+module.exports = ({ prisonerProfileService, elite2Api, systemOauthClient, logError }) => async (req, res) => {
   const { offenderNo } = req.params
 
   try {
+    const systemContext = await systemOauthClient.getClientCredentialsTokens()
```
andrewrlee, line 11: could possibly pass through the users username so it appears as part of the audit?

JamieDraperUK: That's a good point actually. We use client creds in a few places and don't pass the username, and I agree it should.

PR #1084, opened 2020-07-08
Nn 2636 case notes ui tweak

Nimphal:


backend/controllers/prisonerProfile/prisonerCaseNotes.js
```diff
@@ -9,6 +9,7 @@ const {
   MOMENT_TIME,
 } = require('../../../src/dateHelpers')
 const config = require('../../config')
+const { getNamesFromString } = require('../../utils')
```
Nimphal, line 12: Reused string manipulation method we use for POM users with a bit of refactor.

cypress.env
```diff
@@ -9,6 +9,7 @@ KEYWORKER_API_URL=http://localhost:9191/keyworker/
 TOKENVERIFICATION_API_URL=http://localhost:9191/tokenverification/
 NN_ENDPOINT_URL=http://localhost:20200/
 API_DATA_COMPLIANCE_ENDPOINT_URL=http://localhost:9191/datacompliance/
+API_CLIENT_ID=dev
```
Nimphal, line 12: This is an interesting one. When we run the integration tests app locally, it picks up what it can from the end file and sort of fills in the rest. It will take any missing values from the nether (or your .env). So if locally you have the API_CLIENT_ID specified, the auth stubs will always work. Circle doesn't have that, so when the tests log out, they try to access an auth stub that doesn't match match of the client id (defaults to elite2apiclient). 

That's my theory anyway :D

integration-tests/integration/appointments/addAppointment.spec.js
```diff
@@ -91,6 +91,7 @@ context('A user can add an appointment', () => {
     const addAppointmentPage = AddAppointmentPage.verifyOnPage()
     const form = addAppointmentPage.form()
     form.date().type(moment().format('DD/MM/yyyy'))
+    addAppointmentPage.activeDate().click()
```
Nimphal, line 94: Sometimes the datepicker doesn't go away when it's supposed to, covering other fields. We've used this approach successfully in bulk and court appointment, so just doing here as well.

PR #1090, opened 2020-07-09
Nn 2769 print incentive slip

Nimphal:
The Incentive level slip is a shared React component between new-nomis-ui and prisonstaffhub and is used in multiple places. This is a bit of a cheat which means that if you're trying to access the slip from a non-react page, it gets the data from a combination of URL params and api calls.

### DIFF
--- a/integration-tests/integration/prisonerProfile/caseNotesPage.spec.js
+++ b/integration-tests/integration/prisonerProfile/caseNotesPage.spec.js
@@ -6,10 +6,10 @@ const caseNotes = [
   {
     caseNoteId: 12311312,
     offenderIdentifier: 'A1234AA',
-    type: 'KA',
-    typeDescription: 'Key Worker',
-    subType: 'KS',
-    subTypeDescription: 'Key Worker Session',
+    type: 'IEP',
+    typeDescription: 'Incentive Level',
+    subType: 'IEP_WARN',
+    subTypeDescription: 'Incentive Level Warning',
     source: 'INST',
     creationDateTime: '2017-10-31T01:30:00',
     occurrenceDateTime: '2017-10-31T01:30:00',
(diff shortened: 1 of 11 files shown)


integration-tests/integration/prisonerProfile/caseNotesPage.spec.js
```diff
@@ -63,10 +63,30 @@ context('A user can view prisoner case notes', () => {
...
+        'href',
+        '/iep-slip?offenderNo=A12345&offenderName=Smith%2C%20John&location=HMP%20Moorland&casenoteId=12311312&issuedBy=undefined'
+      )
+
+    tableDataRow.caseNotePrintIncentiveLevelSlipLink().click()
+
+    tableDataRow
+      .caseNotePrintIncentiveLevelSlipLink()
+      .invoke('removeAttr', 'target')
+      .click()
```
Nimphal, line 87: Cypress can't cope with multiple windows so this temporarily strips the target="_blank" from the link so it loads in the same window and can be interrogated.

integration-tests/integration/prisonerProfile/caseNotesPage.spec.js
```diff
@@ -63,10 +64,20 @@ context('A user can view prisoner case notes', () => {
...
       .contains('Add more details')
       .should('have.attr', 'href', 'http://localhost:20200/offenders/A12345/case-notes/12311312/amend-case-note')
+    tableDataRow
+      .caseNotePrintIncentiveLevelSlipLink()
+      .contains('Print Incentive Level Slip')
+      .should(
+        'have.attr',
+        'href',
+        '/iep-slip?offenderNo=A12345&offenderName=Smith%2C%20John&location=HMP%20Moorland&casenoteId=12311312&issuedBy=undefined'
+      )
```
Nimphal, line 81: Clicking this link, while successfully loading and checking the page, resulted in a stray print window that can't be closed through cypress which meant the tests took 11m before I noticed.

src/IncentiveLevelSlipContainer/IncentiveLevelSlipContainer.js
```diff
@@ -12,7 +15,32 @@ function IncentiveLevelSlipContainer() {
   const [printed, setPrinted] = useState(false)
 
   useEffect(() => {
-    setIepData(JSON.parse(localStorage.getItem('incentiveLevelSlip')))
+    const setUpData = async () => {
+      const data = localStorage.getItem('incentiveLevelSlip')
+      if (data) {
+        setIepData(JSON.parse(data))
+      } else if (!printed && !iepData) {
```
Nimphal, line 22: Without these checks, the component kept checking, and kept making the request to the api and all of the below. This makes sure it only does the thing when it needs to.

PR #1096, opened 2020-07-10
DCS-519 Adding page to show new arrivals that have not been assigned a unit

andrewrlee:
<kbd>Screenshot 2020-07-10 at 16 46 46
</kbd>


views/covid/notInUnit.njk
```diff
@@ -0,0 +1,106 @@
...
+      ]), rows) %}
+    {% endfor %}
+
+{% block content %}
+    <h1 class="govuk-heading-l">
+        {{ title }}
+    </h1>
+    <p class="govuk-body">
+      These prisoners should all be added to the Reverse Cohorting Unit by creating an alert. If any of these prisoners have been assesed by healthcare 
+      and have to be moved to the Protective Isolation Unit or the Shielding Unit, you first must create that relevant alert for each prisoner.
```
bell-pa, line 57: 'assessed' ?

backend/api/elite2Api.js
```diff
@@ -148,6 +148,9 @@ const elite2ApiFactory = client => {
   const getMovementsIn = (context, agencyId, movementDate) =>
     get(context, `/api/movements/${agencyId}/in/${movementDate}`)
 
+  const getMovementsInBetween = (context, agencyId, params) =>
+    get(context, `/api/movements/${agencyId}/in?${mapToQueryString(params)}`)
```
bell-pa, line 152: between?

backend/services/covidService.js
```diff
@@ -45,5 +48,43 @@ module.exports = {
...
+
+      const offenderAlerts = await elite2Api.getAlerts(context, {
+        agencyId: caseLoadId,
+        offenderNumbers: movements.map(inmate => inmate.offenderNo),
+      })
+
+      const alertsByOffenderNumber = offenderAlerts.reduce((result, alert) => {
+        const bucket = result[alert.offenderNo] || []
+        const newBucket = [...bucket, alert]
+        return { ...result, [alert.offenderNo]: newBucket }
```
bell-pa, line 71: Is this necessary? The arrays and objects aren't immutable. You could just push the new alert onto the bucket. Likewise with spreading result into a new object.

andrewrlee: There's obviously object creation but its pretty explicit.
With linting rules it becomes a bit too verbose and fiddly:
```
        const bucket = result[alert.offenderNo] || []
        bucket.push(alert)
        const newResult = result
        newResult[alert.offenderNo] = bucket
        return newResult
```

backend/api/elite2Api.js
andrewrlee: the endpoint this calls allows getting `movements in` across a range of dates, compared to the method above which just gets the `movements in` on a specific date.

PR #1103, opened 2020-07-14
NN-2792: Fix Quick look tab display

Nimphal:


cypress.env
```diff
@@ -19,4 +19,5 @@ NOTIFY_ENABLED=false
 DISPLAY_RETENTION_LINK=true
 
 REDIS_ENABLED=true
-REDIS_HOST=localhost
\ No newline at end of file
+REDIS_HOST=localhost
+REDIS_PORT=6380
```
Nimphal, line 23: Making sure redis for tests doesn't clash with other redises.

PR #1105, opened 2020-07-15
NN-2726 case note amendments

JamieDraperUK:
Screenshot 2020-07-15 at 11 40 28


Nimphal:
Just from the screenshot, it looks like the date and time column is overlapping with the 'Happened' link. Is that on purpose?

PR #1114, opened 2020-07-16
 NN-2762 Update links to use the new prisoner profiile pages

jsrobertson:


prisonstaffhub-specs/src/test/groovy/uk/gov/justice/digital/hmpps/prisonstaffhub/mockapis/Elite2Api.groovy
```diff
@@ -950,5 +950,187 @@ class Elite2Api extends WireMockRule {
         )
     }
 
+    void stubAlertTypes(response = []) {
```
jsrobertson, line 953: Added these just to get the integration tests working, but plan is to move them to cypress anyway.

PR #1118, opened 2020-07-20
NN-2771: Fix bug when address type is missing

Nimphal:
Also updates a test that requires age calculation. Turns out it was someone's birthday over the weekend.


backend/tests/prisonerVisits.test.js
```diff
@@ -235,7 +240,9 @@ describe('Prisoner visits', () => {
             status: 'Absence',
           },
           {
-            age: '39',
+            age: moment()
+              .diff('1980-07-20', 'years')
+              .toString(),
```
Nimphal, line 245: Birthday is today, so test failed because they've turned 40 :D

PR #1127, opened 2020-07-21
Nn 2741 create alert form

Nimphal:
Transfer Create alert form from React to server-rendered, updating labels and validation in the process.


backend/tests/alert.test.js
```diff
@@ -16,8 +24,15 @@ jest.mock('../logError', () => ({
 config.app.notmEndpointUrl = '//newNomisEndPointUrl/'
 
 describe('alert management', () => {
-  const res = { render: jest.fn(), redirect: jest.fn(), locals: {} }
-  const mockReq = { flash: jest.fn().mockReturnValue([]), originalUrl: '/close-alert/', get: jest.fn(), body: {} }
+  let mockReq
+  let res
```
Nimphal, line 28: The request and response were not resetting after each test, resulting in the test rendering the wrong template even though the right controller was being called.

backend/tests/controllers.test.js
```diff
@@ -1,51 +0,0 @@
-import { factory } from '../controllers/controller'
```
Nimphal, line 1: This was just testing the create alert routes.

integration-tests/integration/alerts/createAlert.spec.js
```diff
@@ -0,0 +1,60 @@
...
+      offenderFullDetails,
+      iepSummary: {},
+      caseNoteSummary: {},
+    })
+    cy.task('stubAlertsForBooking', [])
+    const createAlertPage = CreateAlertPage.verifyOnPage()
+    const form = createAlertPage.form()
+    form.alertType().select('F1')
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(50)
```
Nimphal, line 37: Need to do this to give time for the updated sub type select dropdown to reattach itself to the DOM.

static/js/main.js
```diff
@@ -3,7 +3,8 @@ $(document).ready(function() {
     const disableFutureDates = Boolean($(element).data('disable-future-date'))
     const disablePastDates = Boolean($(element).data('disable-past-date'))
     const maxDate = disableFutureDates ? '0' : undefined
-    const minDate = disablePastDates ? '0' : undefined
+    const minDate = $(element).data('min-date') ? $(element).data('min-date') : disablePastDates ? '0' : undefined
+
```
Nimphal, line 7: Adding support for specifying how many days in the past are allowed

integration-tests/integration/alerts/createAlert.spec.js
JamieDraperUK: Yeah, we need to find a solution for this as it's the same issue with case notes.

PR #1144, opened 2020-07-27
DT-1028 🔧 Update gateway urls

Mjwillis:
🔧 Switch to new Azure Gateway V2 and new domains


jsrobertson:
Suspect the lint failures are related to some of the package updates

Mjwillis:
> looks like you've a space at the end of auth for values-prod?

@petergphillips yup! removed

PR #1149, opened 2020-07-28
NN-2708: Fix case note display, auto populate form with date and time

Nimphal:


sass/application.scss
```diff
@@ -214,3 +214,9 @@ $govuk-page-width: 1170px;
 .case-notes-details {
   flex-basis: 80%;
 }
+
+pre {
+  white-space: pre-wrap;       /* css-3 */
+  white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
+  word-wrap: break-word;       /* Internet Explorer 5.5+ */
+ }
```
Nimphal, line 222: `pre` elements are very permissive so unless you tell them to, they do not constrain content at all.

backend/tests/alert.test.js
```diff
@@ -437,7 +437,7 @@ describe('alert management', () => {
           body: {
             alertType: 'P',
             alertCode: 'PI',
-            effectiveDate: '20/07/2020',
+            effectiveDate: moment().format('DD/MM/YYYY'),
```
Nimphal, line 440: More fixing of hardcoded dates resulting in test failures.

PR #1152, opened 2020-07-28
Nn 2708 fix future time

Nimphal:


backend/controllers/caseNote.js
```diff
@@ -116,14 +116,14 @@ const caseNoteFactory = (elite2Api, caseNotesApi) => {
       })
     }
 
-    if ((hours && parseInt(hours, 10) > 23) || !hours) {
+    if ((hours && !(parseInt(hours, 10) <= 23)) || !hours) {
```
Nimphal, line 119: parseInt returns NaN if letters are passed. NaN > a number and NaN less than a number both return `false`, switching how this checks the time value around means that it will behave correctly both in the case of a NaN and if the value is too high.

backend/controllers/caseNote.js
```diff
@@ -137,6 +137,24 @@ const caseNoteFactory = (elite2Api, caseNotesApi) => {
...
+      if (dateTime > moment().format('YYYY-MM-DDTHH:mm:ss')) {
+        errors.push({
+          text: 'Enter a time which is not in the future',
+          href: '#hours',
+        })
+      }
+    } catch {
+      // Do nothing, will only fail if the date inputs are invalid, which
+      // has been handled by the above validations
+    }
```
Nimphal, line 156: There's no need to do anything in the catch, if the date creation failed it would be because one of the validation errors covered above.

PR #1165, opened 2020-07-31
NN-2727 amend case note

JamieDraperUK:
Screenshot 2020-07-31 at 10 02 52


backend/controllers/amendmentCaseNote.js
```diff
@@ -0,0 +1,91 @@
...
+
+    try {
+      const [caseNote, prisonerDetails] = await Promise.all([
+        caseNotesApi.getCaseNote(res.locals, offenderNo, caseNoteId),
+        elite2Api.getDetails(res.locals, offenderNo),
+      ])
+
+      const { staffId } = req.session.userDetails
+
+      if (Number(caseNote.authorUserId) !== staffId) return res.redirect('/not-found')
```
Nimphal, line 16: Nice!

backend/controllers/amendmentCaseNote.js
```diff
@@ -0,0 +1,91 @@
...
+          text: 'Enter more details to case note',
+        })
+      }
+
+      if (moreDetail && moreDetail.length > 4000) {
+        errors.push({
+          href: '#moreDetail',
+          text: `Enter more details using 4000 characters or less`,
+        })
+      }
```
Nimphal, line 66: Have we decided to leave the validation as is, rather than update it to take into account the length of the current text?

views/amendCaseNote.njk
```diff
@@ -0,0 +1,80 @@
...
+
+    <form action="{{ postAmendment }}" method="post">
+        {{ govukTextarea({
+            name: "moreDetail",
+            id: "moreDetail",
+            value: formValues.moreDetail,
+            errorMessage: errors | findError('moreDetail'),
+            rows: 10,
+            classes: 'govuk-!-width-one-half'
+        }) }}
```
Nimphal, line 68: Absolutely not necessary, but I used the character count govuk component in the create case note template - same thing, but adds a character counter at the bottom. It's not part of the ACs though, so feel free to ignore.

backend/controllers/amendmentCaseNote.js
JamieDraperUK: Yeah, I've added a comment to the ticket. TLDR: It's a lot more complicated because of the way elite to creates it's very own structured data inside a text field, along with text, username dates etc. Because of this, we can't just count the case note text and amendment text because of the extra meta data elite adds.

views/amendCaseNote.njk
JamieDraperUK: Might be worth including once to solve the real characters remaining problem.

PR #1175, opened 2020-08-05
NN-2840 fix minute pre-population 

JamieDraperUK:


backend/controllers/caseNote.js
```diff
@@ -53,8 +53,9 @@ const caseNoteFactory = (elite2Api, caseNotesApi) => {
           subType,
           ...req.body,
           date: currentDateTime.format('DD/MM/YYYY'),
-          hours: currentDateTime.hours(),
-          minutes: currentDateTime.minutes(),
+          hours: currentDateTime.hours().toString(),
+          minutes:
+            currentDateTime.minutes() < 10 ? `0${currentDateTime.minutes()}` : currentDateTime.minutes().toString(),
```
Nimphal, line 58: you could do format('mm') for the minutes and format('H') for the hours as well. Would save converting to string, but not blocking obviously.

JamieDraperUK: I'll do that as it's a lot cleaner.

PR #1177, opened 2020-08-06
Nn 2650 cell move non associations

Nimphal:
Screenshot from 2020-08-06 09-47-05


backend/controllers/cellMove/selectLocation.js
```diff
@@ -51,7 +53,8 @@ module.exports = ({ elite2Api, logError }) => async (req, res) => {
       nonAssociations.nonAssociations.some(
         nonAssociation =>
           nonAssociation.offenderNonAssociation.agencyDescription.toLowerCase() ===
-          prisonerDetails.assignedLivingUnit.agencyName.toLowerCase()
+            prisonerDetails.assignedLivingUnit.agencyName.toLowerCase() &&
```
JamieDraperUK, line 56: Sure it's fine but can offenderNonAssociation and assignedLivingUnit be null?

backend/controllers/cellMove/viewNonAssociations.js
```diff
@@ -0,0 +1,72 @@
...
+
+    return res.render('cellMove/nonAssociations.njk', {
+      nonAssociationsRows,
+      breadcrumbPrisonerName: putLastNameFirst(firstName, lastName),
+      prisonerName: formatName(firstName, lastName),
+      dpsUrl,
+      selectLocationLink: `/prisoner/${offenderNo}/cell-move/select-location`,
+    })
+  } catch (error) {
+    if (error) logError(req.originalUrl, error, serviceUnavailableMessage)
```
JamieDraperUK, line 65: Can error here sometimes be null? I've seen this code elsewhere and wondered the same thing.

Nimphal: Not if it's an API error I don't think, they follow a pattern.

backend/controllers/cellMove/selectLocation.js
Nimphal: Probably :D The API wouldn't return the nonAssociation record without the corresponding offender, and the offender is always serialised the same.

PR #1187, opened 2020-08-10
NN-2744: Add not found page and pages access rules

Nimphal:


backend/index.js
```diff
@@ -41,10 +41,22 @@ app.use(setupAuth({ oauthApi: apis.oauthApi, tokenVerificationApi: apis.tokenVer
...
 })
 
+app.get('/manage-prisoner-whereabouts*', (req, res) => {
+  res.sendFile(path.join(__dirname, '../build/index.html'))
+})
+
+app.get('/offenders/:offenderNo/*', (req, res) => {
+  res.sendFile(path.join(__dirname, '../build/index.html'))
+})
+
```
Nimphal, line 55: Explicitly list the React routes so they do not trigger a server side 404.

backend/index.js
```diff
@@ -41,10 +41,22 @@ app.use(setupAuth({ oauthApi: apis.oauthApi, tokenVerificationApi: apis.tokenVer
...
+app.use((req, res) => {
+  res.redirect(config.app.notmEndpointUrl)
+})
```
Nimphal, line 58: If a page isn't found, redirect back to DPS home.

.circleci/config.yml
```diff
@@ -97,7 +97,7 @@ jobs:
           command: npx cypress install
       - run:
           name: Run the node app.
-          command: npm run start-feature-no-webpack --env=cypress.env
+          command: npm run start-feature --env=cypress.env
```
Nimphal, line 100: Need webpack for tests now, as we point to the react file from than just static.

PR #1189, opened 2020-08-11
NN-2540: Add locality to professional addresses

Nimphal:
Screenshot from 2020-08-11 08-57-11


Nimphal:
No, they're very often missing, but we filter out the missing values in the array, so it won't error out or cause issues.

PR #1206, opened 2020-08-20
NN-2658 Add new risk flags for cell move only

jsrobertson:
image
image


Nimphal (approved):
That is much more sustainably done than what I had in mind, nice work!

PR #1214, opened 2020-08-25
NN-2881 Add CSRA last review date to profile header

jsrobertson:
image

### DIFF
--- a/backend/services/prisonerProfileService.js
+++ b/backend/services/prisonerProfileService.js
@@ -131,7 +131,10 @@ module.exports = ({
     const canViewSocLink = Boolean(isSocUser && socDetails)
 
     const mostRecentAssessment =
-      hasLength(assessments) && assessments.sort((a, b) => b.assessmentDate.localeCompare(a.assessmentDate))[0]
+      hasLength(assessments) &&
+      assessments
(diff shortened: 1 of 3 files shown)


backend/services/prisonerProfileService.js
```diff
@@ -129,6 +130,9 @@ module.exports = ({
 
     const canViewSocLink = Boolean(isSocUser && socDetails)
 
+    const mostRecentAssessment =
+      hasLength(assessments) && assessments.sort((a, b) => b.assessmentDate.localeCompare(a.assessmentDate))[0]
```
Nimphal, line 134: I think you want a more specific assessment here, one with the `assessmentCode: 'CSR'` . Definitely only ones relating to CSR, as re-categorisation assessments will also be returned in this array.

Nimphal: This is what we do in the cell move header:
```const showCsraLink = assessments => {
  return assessments.some(assessment => assessment.assessmentCode === 'CSR' && assessment.assessmentComment)
}```

jsrobertson: Thank you, didn't realise that.

Nimphal (approved):
LGTM

PR #1215, opened 2020-08-26
PAT-1216: Updated URLs for SOC for altered domain name

timharrison-moj:
URL changes for manage-soc-services - updated to use the .hmpps.service.justice.gov.uk domain in all environments and set the enabled flags so that the links will show where the current user has a SOC role in their token.


JamieDraperUK (commented):
lgtm

Nimphal (approved):

PR #1219, opened 2020-08-27
NN-2850 Rename and split incentive and adjudications section on quick look

jsrobertson:
image

Also changed Incentive level casing on details page.


Nimphal (approved):
LGTM

PR #1220, opened 2020-08-27
Nn 2718 csra details page

Nimphal:
Screenshot from 2020-08-27 15-14-12

### DIFF
--- a/backend/controllers/cellMove/viewCellSharingAssessmentDetails.js
+++ b/backend/controllers/cellMove/viewCellSharingAssessmentDetails.js
@@ -1,3 +1,4 @@
+const moment = require('moment')
 const { serviceUnavailableMessage } = require('../../common-messages')
 const { putLastNameFirst, hasLength } = require('../../utils')
 const { getBackLinkData } = require('./cellMoveUtils')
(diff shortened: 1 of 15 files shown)


backend/controllers/cellMove/viewCellSharingAssessmentDetails.js
```diff
@@ -0,0 +1,37 @@
...
+const { putLastNameFirst, hasLength } = require('../../utils')
+const { getBackLinkData } = require('./cellMoveUtils')
+
+module.exports = ({ elite2Api, logError }) => async (req, res) => {
+  const { offenderNo } = req.params
+
+  try {
+    const { firstName, lastName, assignedLivingUnit } = await elite2Api.getDetails(res.locals, offenderNo, true)
+
+    const assessments = await elite2Api.getAssessments(res.locals, { offenderNumbers: [offenderNo], code: 'CSR' })
```
Nimphal, line 12: The offender details endpoint does not return the approval date or the agency Id, so calling an assessment specific endpoint to get all the data. Can be changed to the details page in the future if we update the API.

backend/controllers/cellMove/viewCellSharingAssessmentDetails.js
```diff
@@ -0,0 +1,37 @@
+const moment = require('moment')
+const { serviceUnavailableMessage } = require('../../common-messages')
+const { putLastNameFirst, hasLength } = require('../../utils')
+const { getBackLinkData } = require('./cellMoveUtils')
+
+module.exports = ({ elite2Api, logError }) => async (req, res) => {
+  const { offenderNo } = req.params
+
+  try {
+    const { firstName, lastName, assignedLivingUnit } = await elite2Api.getDetails(res.locals, offenderNo, true)
```
JamieDraperUK, line 10: I think the unspoken rule is to group more than one call into a Promise.all() if possible.

integration-tests/integration/cellMove/viewCellSharingRiskAssessmentDetails.spec.js
```diff
@@ -0,0 +1,77 @@
+const offenderFullDetails = require('../../mockApis/responses/offenderFullDetails')
```
JamieDraperUK, line 1: No fancy Page object? (I like this way better)

JamieDraperUK: Actually, sorry I'm blind. I can see it below 👁️

integration-tests/integration/cellMove/viewCellSharingRiskAssessmentDetails.spec.js
```diff
@@ -0,0 +1,77 @@
...
+          assessorUser: 'TEST_USER',
+        },
+      ],
+    })
+  })
+
+  it('Shows the correct data for non-associations', () => {
+    cy.visit(`/prisoner/${offenderNo}/cell-move/cell-sharing-risk-assessment-details`)
+    const cellSharingRiskAssessmentDetailsPage = CellSharingRiskAssessmentPage.verifyOnPage()
+    cy.get('.govuk-summary-list--no-border').then($section => {
```
JamieDraperUK, line 46: Should this be encapsulated inside the page object?

integration-tests/integration/cellMove/viewCellSharingRiskAssessmentDetails.spec.js
Nimphal: That's just the json response, there's a fancy page object underneath I'm afraid :D

JamieDraperUK (approved):
LGTM

integration-tests/integration/cellMove/viewCellSharingRiskAssessmentDetails.spec.js
Nimphal: we haven't been doing that for summaries as various parts of them are optional. Less so in this one, so probably wouldn't hurt. I'll update when next in the area, I'm sure it won't be long

JamieDraperUK (approved):
LGTM!

PR #1227, opened 2020-09-03
NN-2897 cell filter by attribute

JamieDraperUK:
- Filter by attribute
- Show different default sub location dropdown values depending on the location value
- Added initial cypress test

### DIFF
--- a/backend/controllers/cellMove/selectCell.js
+++ b/backend/controllers/cellMove/selectCell.js
@@ -11,7 +11,7 @@ const extractQueryParameters = query => {
 
   return {
     location: location || 'ALL',
-    attribute: attribute || 'ALL',
+    attribute,
     subLocation,
     locationId,
   }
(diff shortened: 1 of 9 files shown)


backend/controllers/cellMove/selectCell.js
```diff
@@ -28,9 +28,12 @@ module.exports = ({ elite2Api, whereaboutsApi, logError }) => async (req, res) =
 
     if (req.xhr) {
       return res.render('cellMove/partials/subLocationsSelect.njk', {
-        subLocations: locationsData
-          .find(loc => loc.key.toLowerCase() === locationId.toLowerCase())
-          .children.map(loc => ({ text: loc.name, value: loc.key })),
+        subLocations:
+          locationId === 'ALL'
+            ? []
```
Nimphal, line 33: Do you need the 'No location available' dropdown element here? I seem to remember that being a requirement

jsrobertson (approved):

PR #1228, opened 2020-09-03
NN-2898 Add table for available cells

jsrobertson:
image


backend/controllers/cellMove/selectCell.js
```diff
@@ -95,7 +97,15 @@ module.exports = ({ elite2Api, whereaboutsApi, logError }) => async (req, res) =
         nonAssociations && showNonAssociationsLink(nonAssociations, prisonerDetails.assignedLivingUnit),
       showCsraLink: prisonerDetails.assessments && showCsraLink(prisonerDetails.assessments),
       alerts: alertsToShow,
-      cells,
+      cells:
```
Nimphal, line 100: I think the cell description needs to be stripped of the agency id, there's a helper in `utils.js` called `stripAgencyPrefix`. And I also think the location needs to be bold, at least that's what the design has on.

Nimphal (approved):

PR #1230, opened 2020-09-03
Nn 1180 prisoner location history

Nimphal:
Screenshot from 2020-09-03 15-44-46

### DIFF
--- a/backend/controllers/prisonerProfile/prisonerCellHistory.js
+++ b/backend/controllers/prisonerProfile/prisonerCellHistory.js
@@ -37,14 +37,13 @@ module.exports = ({ elite2Api, logError, page = 0 }) => async (req, res) => {
       size: 10000,
     })
 
+    // Collect a list of the unique agencies, to remove
+    // the need to make duplicate API calls
     const agencyData = await getAgencyDetails(cells)
 
     const currentLocation = cells.find(cell => cell.assignmentEndDateTime === undefined)
-    const occupiers =
-      (currentLocation && (await elite2Api.getInmates(res.locals, currentLocation.description, {}))) || []
-
-    const attributes = await elite2Api.getAttributesForLocation(res.locals, currentLocation.livingUnitId)
-    const isSingleOccupancy = attributes && attributes.capacity === 1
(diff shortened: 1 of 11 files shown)


backend/controllers/prisonerProfile/prisonerCellHistory.js
```diff
@@ -0,0 +1,87 @@
...
+} = require('../../config')
+
+module.exports = ({ elite2Api, logError, page = 0 }) => async (req, res) => {
+  const { offenderNo } = req.params
+
+  const getAgencyDetails = async cells => {
+    return Promise.all(
+      cells
+        .map(cell => cell.agencyId)
+        .filter((v, i, a) => a.indexOf(v) === i)
```
Nimphal, line 21: Make the list unique on agency id. Prevents making multiple calls to the API for the same agency detail

backend/controllers/prisonerProfile/prisonerCellHistory.js
```diff
@@ -0,0 +1,87 @@
...
+    if (withoutAgency.includes('RECP')) return 'Reception'
+    if (withoutAgency.includes('CSWAP')) return 'Cell swap'
+    return withoutAgency
+  }
+
+  try {
+    const { bookingId, firstName, lastName } = await elite2Api.getDetails(res.locals, offenderNo)
+    const { content: cells } = await elite2Api.getOffenderCellHistory(res.locals, bookingId, {
+      page,
+      size: 10000,
```
Nimphal, line 37: This is temporary to return all data for now. Pagination will be properly implemented at a later stage.

jsrobertson (approved):

backend/controllers/prisonerProfile/prisonerCellHistory.js
JamieDraperUK: Not sure if it's better but the way we've made list unique is by doing: `Array.from(new Set(data.map(event => event.bookingId)))`

JamieDraperUK (approved):
LGTM 👍

PR #1241, opened 2020-09-10
 NN-2900-non-association-warnings

JamieDraperUK:
image

### DIFF
--- a/sass/components/_alert-flags.scss
+++ b/sass/components/_alert-flags.scss
@@ -225,3 +225,12 @@
     border-color: #00437b;
   }
 }
+
+.non-association {
+  @extend %status-style;
+
+  background: lighten(#D4351C,40);
+  border-color: lighten(#D4351C,40);
+  font-weight: bold;
+  color: #D4351C;
+}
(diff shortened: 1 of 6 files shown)


sass/components/_alert-flags.scss
```diff
@@ -225,3 +225,12 @@
     border-color: #00437b;
   }
 }
+
+.non-association {
```
jsrobertson, line 229: Probably should sit with the other alerts and then be used like `class="alert-status alert-status--non-association"`

JamieDraperUK: I did think that at first but it's not a traditional alert is more of a individual warning.  I put it in here because I wanted to reuse the responsive status css.

JamieDraperUK: What we could do later on is pull the %status-style code out into it's own file then we can use it to create alerts, warnings or anything else that looks similar. What do you think?

jsrobertson: See what you mean.  Leave it for now I reckon then but it might be that we rename the flags so they're not specific to 'alerts', if/when we start adding more and more.

jsrobertson (approved):
LGTM

PR #1256, opened 2020-09-16
NN-2925 show non association warning

JamieDraperUK:
- Show the non association warning when there is an non association on the same residential unit.

image


backend/tests/cellMove/selectCell.test.js
```diff
@@ -35,6 +35,7 @@ describe('Select a cell', () => {
       },
```
JamieDraperUK, line 35: I restructured the test a bit because we've got a lot going on in this page and the tests started to look messy.

Nimphal (approved):

PR #1262, opened 2020-09-17
NN-2588: Enable cell move button for users with the role

Nimphal:
Make the Change cell button appear conditional on the user having the 'CELL_MOVE' role. The role currently only exists on dev.


JamieDraperUK (approved):
LGTM

PR #1263, opened 2020-09-17
Nn 2716 fix category warning

Nimphal:
Previously whether the category warning showed up was dependent on there being alerts, the two were wrongly tied together. This decouples the logic and ensures the category warning shows when the offender is cat A and there are other offenders in the cell.


JamieDraperUK (approved):
lgtm

PR #1265, opened 2020-09-17
NN-2716: Update behaviour for LGB alert

Nimphal:
Previously we were showing the LGB alert regardless of the sexualities involved, it was just the wording that changed. This updates it so that we only show the alert if there are non-straight sexualities involved, otherwise do not include the alert in the risks at all.


JamieDraperUK (approved):
lgtm!

PR #1267, opened 2020-09-17
Nn 2905 cell move access restriction

Nimphal:
If a user is trying to access cell move for an offender that's not in one of their case loads, take them to the page not found page instead.


JamieDraperUK (approved):
LGTM

PR #1270, opened 2020-09-18
NN-2932: Specify date format for query string so moment does not have to guess

Nimphal:
How this has worked at all is beyond me. JS magic.


JamieDraperUK (approved):
LGTM

PR #1271, opened 2020-09-21
Nn 2923 csrf protection

Nimphal:
This enables CSRF protection on the server rendered pages. It also breaks out the routing into separate concerns so we can handle React and server-side appropriately.


JamieDraperUK (approved):
Amazing work!!

PR #1273, opened 2020-09-22
NN-2677: Convert Edit alert integration tests to cypress

Nimphal:
Also deletes a bunch of mocks that are no longer used in the groovy tests.


JamieDraperUK (approved):
LGTM!

PR #1276, opened 2020-09-22
Nn 2693 court bookings spec

Nimphal:
Convert video link court bookings integration test from groovy to cypress


JamieDraperUK (approved):
lgtm

PR #1282, opened 2020-09-24
Nn 2691 whereabouts search cypress test

JamieDraperUK:


integration-tests/mockApis/whereabouts.js
```diff
@@ -294,4 +294,24 @@ module.exports = {
         jsonBody: response,
       },
     }),
+  stubLocationGroups: locationGroups =>
```
Nimphal, line 297: There's a stub called `stubGroups` which I think might be doing something very similar? Not that I mind :D

Nimphal (approved):
LGTM!

integration-tests/mockApis/whereabouts.js
JamieDraperUK: Ah you might be right! I'll fix it on the next PR

PR #1288, opened 2020-09-28
NN-2948 Update prisoner information UI to be better across all screen sizes

jsrobertson:
image

image


JamieDraperUK (approved):
Nice one 👍

PR #1291, opened 2020-10-05
Nn 2936 update adjudication history

JamieDraperUK:
- Rewrite adjudication history page using server side tech
- Remove adjudication react code and groovy tests

image


jsrobertson (approved):

backend/routes.js
```diff
@@ -218,6 +220,8 @@ const setup = ({
     amendCaseNNoteRouter({ elite2Api, caseNotesApi, logError })
   )
 
+  router.use('/offenders/:offenderNo/adjudications', adjudicationsHistoryRouter(elite2Api, logError))
```
jsrobertson, line 223: Most prisoner stuff is now at /prisoner/.  Do we plan going to change this to make it consistent?

jsrobertson: Small thing which threw me when doing a find was we've got both adjudicationsHistoryRouter and adjudicationHistoryRouter (no s).

PR #1293, opened 2020-10-06
NN-2795 Add possessive name nunjucks filter

jsrobertson:
Name ending with an S:
image

Name NOT ending with an S:
image


backend/controllers/appointments/confirmAppointment.js
```diff
@@ -175,14 +175,10 @@ const confirmAppointmentFactory = ({ elite2Api, appointmentsService, logError })
...
         }
-        const formattedName = sanitizer.sanitize(formatName(firstName, lastName))
-
-        const prisonerName =
-          formattedName && formattedName[formattedName.length - 1] !== 's' ? [formattedName, 's'] : [formattedName]
 
         res.render('confirmAppointments.njk', {
-          titleHtml: `${prisonerName[0]}&rsquo;${prisonerName[1] || ''} appointment has been added`,
           addAppointmentsLink: `/offenders/${offenderNo}/add-appointment`,
+          prisonerName: sanitizer.sanitize(formatName(firstName, lastName)),
```
jsrobertson, line 181: Not familar with this @JamieDraperUK. Any reason we do this `sanitizer` here?

JamieDraperUK: It's to stop XSS attacks given that the name presented in the view was unsafe. An attacker could change an offenders name to `<script> alert('do something malicious') </script> Bob Doe` Not sure if there is an easy way to solve it. I tried using | safe but because the older code did some string building in the view it didn't work very well.

backend/utils.js
```diff
@@ -252,7 +252,11 @@ const times = number => func => {
   return iter(0)
 }
 
-const possessive = string => (string.toLowerCase().endsWith('s') ? '&rsquo;' : '&rsquo;s')
+const possessive = string => {
+  if (!string) return ''
+
+  return `${string}${string.toLowerCase().endsWith('s') ? '’' : '’s'}`
+}
```
Nimphal, line 259: ooh, does this do the right curly apostrophe in the view?

backend/controllers/adjudicationHistory.js
```diff
@@ -0,0 +1,110 @@
+const moment = require('moment')
```
Nimphal, line 1: Is this file supposed to be part of this PR?

jsrobertson: Nope :D There was a merge conflict when I pulled in from main so looks like it got re added by accident.  Removing, cheers!

backend/utils.js
jsrobertson: Yeah it's the same as `&rsquo;`

Nimphal (approved):

PR #1296, opened 2020-10-06
DT-1033: global search call to prisonAPI not needed when prisoner is …

SimonMitchellMOJ:
…'OUT' as prison-offender-search now has locationDescription which holds this data,


backend/api/offenderSearchApi.js
```diff
@@ -8,19 +8,31 @@ const offenderSearchApiFactory = client => {
...
+    data.map(
+      ({
+        uiId,
+        prisonerNumber,
+        bookingId,
+        firstName,
+        lastName,
+        dateOfBirth,
+        prisonId,
+        locationDescription,
```
Nimphal, line 20: Does this rename prisonName to locationDescription?

petergphillips: prisoner offender search has been changed to have a new field - `locationDescription`.  This code reads that new field and maps it to `latestLocation`

petergphillips (approved):
LGTM

backend/api/offenderSearchApi.js
Nimphal: prisonName seems to have dropped our of the list above then. Could very well be intentional, just raising in case not.

backend/controllers/globalSearch.js
```diff
@@ -3,7 +3,7 @@ const shortid = require('shortid')
 const log = require('../log')
 const { isPrisonerIdentifier } = require('../utils')
 
-const globalSearchFactory = (elite2Api, globalSearchApi) => {
+const globalSearchFactory = globalSearchApi => {
   const searchByOffender = (context, offenderNo, gender, location, dateOfBirth) =>
     // offenderNo can be removed once we have fully switched to prisoner offender search instead, i.e. the env variable
     // OFFENDER_SEARCH_API_ENABLED has been removed
```
Nimphal, line 9: Has this comment been addressed?

SimonMitchellMOJ: wanted to get the change working before addressing comment - will pick up next - thanks

backend/api/offenderSearchApi.js
petergphillips: yes that's because we were reading from the `prisonName` but are name reading from `locationDescription` instead.  We still map to the `latestLocation` in both instances

PR #1304, opened 2020-10-08
Nn 2938 replace node sass

JamieDraperUK:
Node-sass will be marked as deprecated soon, so we decided to move away from it sooner rather than later.  

https://github.com/sass/node-sass-middleware/issues/134#issuecomment-599245988

The recommendation is to go with dart, and this seems to be the default compiler on the sass-lang website.

At the moment there isn't a node middleware for it that would take requests, load, compile then stream sass files.  Previously node-sass-middleware took care of this for us. 

With that in mind, we've just opted to integrate with the cli through npm run commands. So to compile everything `npm run compile-sass` or `npm build`.  You can also just run `npm run compile-sass-watch` if you like.

In theory, this should be faster because the pre-processing only happens once.


package-lock.json
```diff
@@ -13226,10 +13281,24 @@
         "walker": "~1.0.5"
       }
     },
+    "sanitizer": {
```
Nimphal, line 13284: We didn't last long without sanitizer did we :D

JamieDraperUK: Opps

Nimphal (approved):
Awesome work! 

LGTM!

Nimphal (approved):

PR #1305, opened 2020-10-08
NN-2979: Update to use the last recorded charge for an adjudication

Nimphal:
Currently we show the first charge for an adjudication, this is to change it so we show the final one.


JamieDraperUK (approved):
Nice one!

PR #1324, opened 2020-10-15
NN-2984 move in-reception out of react

JamieDraperUK:
image


sass/application.scss
```diff
@@ -20,7 +20,7 @@ $govuk-page-width: 1170px;
 @import 'components/covid/dashboard';
 @import 'components/prisonerSearch/prisoner-search';
 @import 'components/prisonerSearch/prisoner-panel';
-
+@import 'components/results-table';
```
Nimphal, line 23: you should be able to just wrap your table in a div with https://github.com/ministryofjustice/prisonstaffhub/blob/main/sass/components/_establishment-roll.scss#L9 this class and it should add some of it, in fact it should probably be updated with what you've added.

Nimphal (approved):
LGTM

sass/application.scss
JamieDraperUK: I've opted to pull the styles for general results table into a reusable .results-table class. In the hopes it makes thing easier going forward and reduce duplication.

PR #1326, opened 2020-10-15
NN-2986: Move Currently out out of React

Nimphal:
Can't delete the react yet as it is used by the Total currently out page as well.

### DIFF
--- a/backend/controllers/establishmentRoll/currentlyOut.js
+++ b/backend/controllers/establishmentRoll/currentlyOut.js
@@ -1,0 +1,46 @@
+const moment = require('moment')
+
+const {
+  app: { notmEndpointUrl: dpsUrl },
+} = require('../../config')
+
+const { alertFlagLabels } = require('../../shared/alertFlagValues')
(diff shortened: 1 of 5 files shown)


backend/controllers/establishmentRoll/currentlyOut.js
```diff
@@ -0,0 +1,46 @@
...
+        }
+      })
+
+    return res.render('establishmentRoll/currentlyOut.njk', {
+      results,
+      livingUnitName: response.location,
+      notmUrl: dpsUrl,
+    })
+  } catch (error) {
+    if (error) logError(req.originalUrl, error, 'Failed to load en route page')
```
JamieDraperUK, line 39: Is it worth adding a unit test for this bit in future PRs?

JamieDraperUK (approved):
LGTM

backend/controllers/establishmentRoll/currentlyOut.js
Nimphal: It definitely won't hurt, I'll add it with the next one.

PR #1330, opened 2020-10-16
Nn 2987 in today

Nimphal:
Discovered we were missing the category flags on the other pages we moved as well.


JamieDraperUK (approved):
LGTM

PR #1332, opened 2020-10-16
NN-2887 Fix printed grid view for prisoner search

jsrobertson:


views/prisonerSearch/partials/prisonerSearchGrid.njk
```diff
@@ -1,9 +1,9 @@
 {% from "../macros/prisonerPanel.njk" import prisonerPanel %}
 
 <div class="prisoner-search__results-grid">
-  {% for prisoner in results %}  
+  {% for prisoner in results -%}
```
Nimphal, line 4: oooh, what does this do?

Nimphal (approved):
LGTM

views/prisonerSearch/partials/prisonerSearchGrid.njk
jsrobertson: https://mozilla.github.io/nunjucks/templating.html#whitespace-control

On this occasion when using `display: inline-block;` and `width: 25%` it was causing the fourth item to wrap onto the next line because there was whitespace in between each element so doing the above stripped the whitespace out and did the trick.

PR #1334, opened 2020-10-19
NN-3000: use image style and set class name

jamesak-oncloud9:
use image style and set class name


sass/components/_results-table.scss
```diff
@@ -1,8 +1,8 @@
-.results-table {
+.f {
```
JamieDraperUK, line 1: interesting class name.

jamesak-oncloud9: oh no it was a typo, reverted like it was :)

JamieDraperUK (approved):
LGTM!

PR #1340, opened 2020-10-26
NN-3003 replace pagination service

JamieDraperUK:


backend/tests/paginationService.test.js
```diff
@@ -1,48 +1,182 @@
 const service = require('../services/paginationService')
 
-describe('Pagination service', () => {
-  it('should return pages, previous, next and results when more than one page', () => {
-    const response = service.getPagination(100, 0, 20, new URL('http://localhost/'))
+describe('Pagination service 2', () => {
```
Nimphal, line 3: Definitely non-consequential, but I think this shouldn't have the 2 in it?

Nimphal (commented):
Does this also work on the visits history page? I think that's the only one we have which uses the new pagination approach.

Nimphal (approved):
LGTM!

JamieDraperUK:
Good question, I'll check visits history page now @Nimphal

PR #1358, opened 2020-11-03
DCS-680 Removing axios mocking library and just using nock instead

andrewrlee:


JamieDraperUK (approved):
LGTM

andrewrlee:
The test fail as something is different in book video link and don't really have time to fix

PR #1359, opened 2020-11-03
NN-2961: show phase name banner

jamesak-oncloud9:


backend/config.js
```diff
@@ -1,3 +1,11 @@
+const phaseName = () => {
+  const nameFromEnv = process.env.SYSTEM_PHASE
+  if (nameFromEnv === 'DEV' || nameFromEnv === 'PRE-PRODUCTION') {
```
JamieDraperUK, line 3: Not sure we need to be this strict. I'd probably just return the name.  

From a javascript standpoint I'd might write the following as:

`
const validPhaseNames = ['DEV', 'PRE_PRODUCTION']
return null ?? validPhaseNames.includes(process.env.SYSTEM_PHASE)
`

backend/config.js
```diff
@@ -113,4 +121,6 @@ module.exports = {
       },
     },
   },
+
+  phaseName: phaseName(),
```
JamieDraperUK, line 125: I'd just expose the name without checks. 
`phaseName: process.env.SYSTEM_PHASE`

backend/index.js
```diff
@@ -32,6 +32,16 @@ app.set('view engine', 'njk')
 
 nunjucksSetup(app, path)
 
+app.use((req, res, next) => {
```
JamieDraperUK, line 35: Just so you know we normally merged objects using the spread operator. 
So this:
`bject.assign(options, { phaseName: config.phaseName })`

Becomes:
`    const optionsWithPhaseName = {...options, phaseName: config.phaseName }`

On another note there is established pattern for this sort of thing already, essentially it's a bit of middleware that lives in setUpRoutes.js and places data into local state, namely res.locals.

I'd put the following in setUpRoutes.js
`  router.use(req, res, next) => {
    res.locals = {
      ...res.locals,
       phaseName: config.phaseName 
    }
    next()
  })`

res.locals will be available to the views automatically.

jamesak-oncloud9: how about simply `app.locals.phaseName = config.phaseName` this seems more natural to what needs to be done as the value never changes once the application is running.

backend/index.js
```diff
@@ -32,6 +32,9 @@ app.set('view engine', 'njk')
 
 nunjucksSetup(app, path)
 
+app.locals.phaseName = config.phaseName
```
JamieDraperUK, line 35: Could we put this into it's own setup function? 

Something along the lines of: 
`app.use(setupPhase??())`   

That way it'll be consistent with everything else in this file.

views/partials/header.njk
```diff
@@ -33,3 +35,17 @@
...
+
+ {% if phaseName %}
+    <div class="outer-phase-banner">
+
+        {{ govukPhaseBanner({
+            tag: {
+                text:  phaseName,
+                classes: phaseNameColour
+            },
+            html: 'This test version of DPS contains real data which may be up to two weeks old.'
```
JamieDraperUK, line 47: Is this messaged also shown in dev?

jamesak-oncloud9: yes same message for both DEV and PRE-PROD

JamieDraperUK (approved):
:+1:

PR #1364, opened 2020-11-04
Removing hsts dependency

andrewrlee:
This is part of helmet so doesn't need to be registered as separate middleware


JamieDraperUK (approved):
Nice one :+1:

PR #1368, opened 2020-11-05
DCS-667 PSH redirect to bvl service

GurnankCheema:


backend/setupRedirects.js
```diff
@@ -22,5 +22,25 @@ module.exports = () => {
     res.redirect(301, `/prisoner/${req.params.offenderNo}/incentive-level-details`)
   )
 
+  if (config.app.featureFlags.redirectToBookingVideoLinkEnabled) {
+    const bvlRoutes = [
```
andrewrlee, line 26: So the idea would be to move this into the new `setupBvlRoutes.js`
```js
if (enabled) {
  // just add redirect for `/videolink` on its own
} else {
 // register all routes
}
```
When we remove vlb entirely from here we might want to semi-permanently add the `/videolink` redirect here.

GurnankCheema: i see.

backend/tests/setupBvlRoutes.test.js
```diff
@@ -0,0 +1,41 @@
...
+    expect(path).toBe('/videolink')
+
+    const res = {
+      redirect: jest.fn(),
+    }
+
+    handler(null, res)
+
+    expect(res.redirect).toHaveBeenCalledWith(config.apis.bookVideoLink.url)
+  })
```
andrewrlee, line 33: This is a bit gnarly but think it's probably acceptable given we'll be removing and simplifying the video link booking code and just leaving a simple redirect that we can but in with the standard redirects

JamieDraperUK (dismissed):
LGTM

JamieDraperUK (approved):

PR #1377, opened 2020-11-16
PAT-1248 and PAT-1291 Create Pathfinder Reader roles

steverendell:
Replaced existing obsolete unused roles


JamieDraperUK (approved):
:+1:

PR #1382, opened 2020-11-18
NN-3050: set referrer policy

jamesak-oncloud9:


backend/index.js
```diff
@@ -35,6 +36,12 @@ app.set('view engine', 'njk')
 nunjucksSetup(app)
 phaseNameSetup(app, config)
 
+app.use(
+  helmet({
```
jamesak-oncloud9, line 40: this could go into a setup file?

jsrobertson: We already have helmet in `setupWebSecurity.js`

jsrobertson: https://github.com/ministryofjustice/prisonstaffhub/blob/main/backend/setupWebSecurity.js#L15

jsrobertson (approved):

PR #1383, opened 2020-11-18
NN-3033 Fix from and to dates on location history location titles

jsrobertson:
Cell movements are pre sorted in descending order before grouping, so the `fromDateString` and `toDateString` were the wrong way round.  Tests only had 1 record before so passed regardless. 😳 🙈 

Dates now correct as shown in screenshot below:
**From =** earliest move in date
**To =** latest move out date
image


JamieDraperUK (approved):
:+1:

PR #1384, opened 2020-11-19
PAT-951: Update SOC referral button text

connormaglynn:
Changed the wording in the service to 'Add' instead of Refer.

Also had to remove the async methods in the tests to appease the linter.
It's highlighted a failing test.

It seems we want the following error to show for the test
image

but instead get
image

This is toggled by line 196 in the ./backend/controllers/prisonersProfile/prisonerQuicklook.js
` personalDetailsSectionError: Boolean(prisonerDataResponse.error && prisonerProfileDataResponse.error),`

Can't seem to make out where prisonerProfileDataResponse.error is set so think the issue is around that?

Solutions:
- Make this an || operator (though I don't think prisonerProfileDataResponse.error is ever set, may just be me being blind)
- Remove prisonerProfileDataResponse.error from the check (though not sure if this is the desire functionality either)
- Add some logic in the retrieval of the prisonerProfileDataResponse (if not there) to set the error, did try appending it to the return object as true to try out, that didn't work so not entirely sure on that one

### DIFF
--- a/integration-tests/integration/prisonerProfile/prisonerQuickLook.spec.js
+++ b/integration-tests/integration/prisonerProfile/prisonerQuickLook.spec.js
@@ -184,7 +184,7 @@ context('Prisoner quick look data retrieval errors', () => {
       })
   })
 
-  it('Should display the appropriate message when there was an error requesting personal details', async () => {
+  it('Should display the appropriate message when there was an error requesting personal details', () => {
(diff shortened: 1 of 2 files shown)


petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #1390, opened 2020-11-23
NN-2462: show damage obligations when over zero

jamesak-oncloud9:
Finance - quick look section


backend/controllers/prisonerProfile/prisonerQuickLook.js
```diff
@@ -138,15 +138,35 @@ module.exports = ({ prisonerProfileService, prisonApi, logError }) => async (req
...
-            value: formatCurrency((balanceData && balanceData.savings) || 0, balanceData && balanceData.currency),
+            visible: true,
+            html: `<a href='#'>${formatCurrency(
+              (balanceData && balanceData.savings) || 0,
+              balanceData && balanceData.currency
+            )}</a>`,
+          },
+          {
+            label: 'Damage obligations',
+            visible: balanceData && balanceData?.damageObligations > 0,
```
jsrobertson, line 165: `balanceData &&` is not needed if doing `balanceData?.'

integration-tests/integration/prisonerProfile/Finances.spec.js
```diff
@@ -0,0 +1,218 @@
+const prisonerQuickLookPage = require('../../pages/prisonerProfile/prisonerQuickLookPage')
```
jsrobertson, line 1: Contents of this file should probably be within `prisonerQuickLookSpec.js` and under a `describe('Finances section'...`

backend/controllers/prisonerProfile/prisonerQuickLook.js
```diff
@@ -138,15 +138,35 @@ module.exports = ({ prisonerProfileService, prisonApi, logError }) => async (req
         balanceDetails: [
           {
             label: 'Spends',
-            value: formatCurrency((balanceData && balanceData.spends) || 0, balanceData && balanceData.currency),
+            visible: true,
+            html: `<a href='#'>${formatCurrency(
```
jsrobertson, line 142: The urls for each page are within the ACs for each of the individual page.  However I think we should add these as we go so we're not linking to a non existant page.  At the moment, damage obligations is the only one that exists at `prisoner/:offenderNo/prisoner-finance-details/damage-obligations`

JamieDraperUK (approved):

PR #1392, opened 2020-11-24
DCS-697 Call new endpoint to book video link appointment

andrewrlee:
Having to update both prison and court side until we are confident with the book service being in prod


GurnankCheema (approved):
LGTM

PR #1394, opened 2020-11-24
DCS-677 Enabling feature flag for BVL redirect

andrewrlee:
Also moving flag out of feature flag section as this was leading to strange react errors (This flag isn't used by react anyway)


jsrobertson (approved):
LGTM @andrewrlee

PR #1395, opened 2020-11-24
NN-3034: show new fields plus test

jamesak-oncloud9:
Screenshot 2020-11-30 at 10 24 02


backend/api/prisonApi.js
```diff
@@ -248,6 +248,9 @@ const prisonApiFactory = client => {
   const getCaseNoteSummaryByTypes = (context, params) =>
     get(context, `/api/case-notes/summary?${mapToQueryString(params)}`)
 
+  const getCaseNote = (context, offenderId, caseNoteId) =>
```
JamieDraperUK, line 251: Is there a equivalent in the caseNoteApi.js?

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -12,7 +12,37 @@ const {
   app: { notmEndpointUrl: dpsUrl },
 } = require('../../config')
 
-module.exports = ({ prisonApi, logError }) => async (req, res) => {
+const fetchStaffName = (context, staffId, prisonApi) => {
+  return prisonApi
+    .getStaffDetails(context, staffId)
+    .then(staff => formatName(staff.firstName, staff.lastName))
+    .catch('No staff name found')
```
JamieDraperUK, line 19: This is never used.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -12,7 +12,37 @@ const {
   app: { notmEndpointUrl: dpsUrl },
 } = require('../../config')
 
-module.exports = ({ prisonApi, logError }) => async (req, res) => {
+const fetchStaffName = (context, staffId, prisonApi) => {
```
JamieDraperUK, line 15: The curly braces and return isn't needed. You can inline it.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -12,7 +12,37 @@ const {
...
+    .then(caseNoteTypes => caseNoteTypes.find(type => type.code === 'MOVED_CELL'))
+    .then(cellMoveTypes => {
+      return cellMoveTypes?.subCodes.map(subType => ({
+        value: subType.code,
+        text: subType.description,
+      }))
+    })
+    .then(cellMoveReasonRadioValues => cellMoveReasonRadioValues.find(record => record.value === assignmentReasonCode))
+    .then(assignmentReason => assignmentReason.text)
+    .catch('No reason description found')
```
JamieDraperUK, line 34: This isn't used.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -26,11 +56,26 @@ module.exports = ({ prisonApi, logError }) => async (req, res) => {
     ])
 
     const userCaseLoadIds = userCaseLoads.map(caseLoad => caseLoad.caseLoadId)
-
     const { bookingId, firstName, lastName } = prisonerDetails
-
     const currentPrisonerDetails = locationHistory.find(record => record.bookingId === bookingId) || {}
 
+    const movementMadeByName = currentPrisonerDetails.movementMadeBy
+      ? await fetchStaffName(res.locals, currentPrisonerDetails.movementMadeBy, prisonApi)
```
JamieDraperUK, line 63: This will throw an exception if it fails. With error containing No staff name found. Is that expected?

backend/controllers/prisonerProfile/prisonerLocationHistory.js
JamieDraperUK: This code ignores the original error and replaces it with: **No staff name found** This could be okay but looking at the code lower down you call it with an **await** This will throw an Error that is caught by the try catch block resulting in a string as the error parameter. It would also be passed into the logError function resulting more errors thrown due to that function expecting an Error object.

backend/tests/prisonerLocationHistory.test.js
```diff
@@ -5,6 +5,8 @@ describe('Prisoner location sharing history', () => {
   const offenderNo = 'ABC123'
```
JamieDraperUK, line 5: Missing unit tests to cover the logic that generates: **No staff name found** and the other custom error wrappers.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -12,7 +12,37 @@ const {
...
+  return caseNotesApi
+    .getCaseNoteTypes(context)
+    .then(caseNoteTypes => caseNoteTypes.find(type => type.code === 'MOVED_CELL'))
+    .then(cellMoveTypes => {
+      return cellMoveTypes?.subCodes.map(subType => ({
+        value: subType.code,
+        text: subType.description,
+      }))
+    })
+    .then(cellMoveReasonRadioValues => cellMoveReasonRadioValues.find(record => record.value === assignmentReasonCode))
```
JamieDraperUK, line 32: Could the result of the then on line 26 result in an undefined value? This could mean that cellMoveReasonRadioValues is undefined.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
JamieDraperUK, line 34: Oh I see. If `cellMoveReasonRadioValues` or `assignmentReason` isn't defined then it will throw an error but your catching it and translating it here. But losing the rest of the Error info.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -26,11 +56,26 @@ module.exports = ({ prisonApi, logError }) => async (req, res) => {
...
 
     const userCaseLoadIds = userCaseLoads.map(caseLoad => caseLoad.caseLoadId)
-
     const { bookingId, firstName, lastName } = prisonerDetails
-
     const currentPrisonerDetails = locationHistory.find(record => record.bookingId === bookingId) || {}
 
+    const movementMadeByName = currentPrisonerDetails.movementMadeBy
+      ? await fetchStaffName(res.locals, currentPrisonerDetails.movementMadeBy, prisonApi)
+      : ''
```
JamieDraperUK, line 64: Should the default be blank?

backend/api/prisonApi.js
jamesak-oncloud9: seems there is  @GetMapping("/{offenderIdentifier}/{caseNoteIdentifier}") shall I use that instead?

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -12,7 +12,37 @@ const {
...
+    .then(assignmentReason => assignmentReason.text)
+    .catch('No reason description found')
+}
+
+const fetchWhatHappened = (context, offenderNo, bookingId, bedAssignmentHistorySequence, prisonApi, whereaboutsApi) => {
+  return whereaboutsApi
+    .getCellMoveReason(context, bookingId, bedAssignmentHistorySequence)
+    .then(cellMoveReason => prisonApi.getCaseNote(context, offenderNo, cellMoveReason.cellMoveReason.caseNoteId))
+    .then(caseNote => caseNote.text)
+    .catch(err => 'Not entered')
```
JamieDraperUK, line 42: This seems different to how we handle these types of errors elsewhere what was the reason for doing it like this?

backend/controllers/prisonerProfile/prisonerLocationHistory.js
jamesak-oncloud9: I agree this looks wrong as it is now it should be `.catch(err => 'No staff name found')` and what I intend from this: is that if any errors occur during the fetching process then the staff name would be set to 'No staff name found', does that sound correct?

backend/controllers/prisonerProfile/prisonerLocationHistory.js
jamesak-oncloud9: yes it should be `.catch(err => 'No reason description found') ` as default value when an error occurs, agree?

backend/controllers/prisonerProfile/prisonerLocationHistory.js
jamesak-oncloud9: if it fails it should return `'No staff name found'` I'll write a test for this

backend/tests/prisonerLocationHistory.test.js
jamesak-oncloud9: there is a test for whatHappened will do the other two as well

backend/controllers/prisonerProfile/prisonerLocationHistory.js
jamesak-oncloud9: yes changed it to  `.catch(err => 'No reason description found')` to return a user-friendly message, is that okay or should log more info about the error?

jamesak-oncloud9: I understand that these errors only would occur if there was a programming error in the code, is that right?

```
const fetchStaffName = (context, staffId, prisonApi) =>
  prisonApi
    .getStaffDetails(context, staffId)
    .then(staff => formatName(staff.firstName, staff.lastName))
    .catch(err => 'No staff name found')

const fetchReasonDescription = (context, assignmentReasonCode, caseNotesApi) =>
  caseNotesApi
    .getCaseNoteTypes(context)
    .then(caseNoteTypes => caseNoteTypes.find(type => type.code === 'MOVED_CELL'))
    .then(cellMoveTypes => {
      return cellMoveTypes?.subCodes.map(subType => ({
        value: subType.code,
        text: subType.description,
      }))
    })
    .then(cellMoveReasonRadioValues => cellMoveReasonRadioValues.find(record => record.value === assignmentReasonCode))
    .then(assignmentReason => assignmentReason.text)
    .catch(err => 'No reason description found')

const fetchWhatHappened = (context, offenderNo, bookingId, bedAssignmentHistorySequence, prisonApi, whereaboutsApi) =>
  whereaboutsApi
    .getCellMoveReason(context, bookingId, bedAssignmentHistorySequence)
    .then(cellMoveReason => prisonApi.getCaseNote(context, offenderNo, cellMoveReason.cellMoveReason.caseNoteId))
    .then(caseNote => caseNote.text)
    .catch(err => 'Not entered')
```

jamesak-oncloud9: or networking issues of course

backend/controllers/prisonerProfile/prisonerLocationHistory.js
JamieDraperUK: It will throw a new error when used with await.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
JamieDraperUK: That is correct. I misunderstood how it worked. However as we've discussed it's probably best to just let the error bubble up so that it gets logged correctly.

backend/api/prisonApi.js
```diff
@@ -248,6 +248,9 @@ const prisonApiFactory = client => {
...
+    get(context, `/api/offenders/${offenderId}/case-notes/${caseNoteId}`)
```
JamieDraperUK, line 252: This needs to go into the caseNotesApi.js module

JamieDraperUK: Check to see if there is already an existing method in the caseNotesApi.js module.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -12,7 +12,43 @@ const {
   app: { notmEndpointUrl: dpsUrl },
 } = require('../../config')
 
-module.exports = ({ prisonApi, logError }) => async (req, res) => {
+const fetchStaffName = (context, staffId, prisonApi) =>
+  prisonApi.getStaffDetails(context, staffId).then(staff => formatName(staff.firstName, staff.lastName))
+
+const fetchReasonDescription = (context, assignmentReasonCode, caseNotesApi) => {
+  if (assignmentReasonCode) {
```
JamieDraperUK, line 19: This seems strange. Sometimes it returns a promise and other times a string. Might be worth doing the checkout outside of this method.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -12,7 +12,43 @@ const {
...
+    return caseNotesApi
+      .getCaseNoteTypes(context)
+      .then(caseNoteTypes => caseNoteTypes.find(type => type.code === 'MOVED_CELL'))
+      .then(cellMoveTypes => {
+        return cellMoveTypes?.subCodes.map(subType => ({
```
JamieDraperUK, line 24: Could inline this.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -12,7 +12,43 @@ const {
...
+  bookingId,
+  bedAssignmentHistorySequence,
+  caseNotesApi,
+  whereaboutsApi
+) =>
+  whereaboutsApi
+    .getCellMoveReason(context, bookingId, bedAssignmentHistorySequence)
+    .then(cellMoveReason => caseNotesApi.getCaseNote(context, offenderNo, cellMoveReason.cellMoveReason.caseNoteId))
+    .then(caseNote => caseNote.text)
+    .catch(err => 'Not entered')
```
JamieDraperUK, line 49: This will show false information to the user if it fails for reasons other than a 404.

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -12,7 +12,43 @@ const {
   app: { notmEndpointUrl: dpsUrl },
 } = require('../../config')
 
-module.exports = ({ prisonApi, logError }) => async (req, res) => {
+const fetchStaffName = (context, staffId, prisonApi) =>
+  prisonApi.getStaffDetails(context, staffId).then(staff => formatName(staff.firstName, staff.lastName))
+
+const fetchReasonDescription = (context, assignmentReasonCode, caseNotesApi) => {
+  if (assignmentReasonCode) {
+    return caseNotesApi
```
JamieDraperUK, line 20: Be really interested to see what the errors look like when they fail on the mapping part when making an API call.

backend/tests/prisonerLocationHistory.test.js
```diff
@@ -28,7 +30,26 @@ describe('Prisoner location sharing history', () => {
     prisonApi.userCaseLoads = jest.fn().mockResolvedValue([])
     prisonApi.getPrisonerDetail = jest.fn()
 
-    controller = prisonerLocationHistory({ prisonApi, logError })
+    prisonApi.getStaffDetails = jest.fn().mockResolvedValue({ firstName: 'Joe', lastName: 'Bloggs' })
+    caseNotesApi.getCaseNote = jest.fn().mockResolvedValue({ text: 'Some details regarding what happened' })
+
+    const caseNotesTypes = [
```
JamieDraperUK, line 36: These are duplicated further down. If you pull them out of the describe you can use them everywhere in the module

backend/controllers/prisonerProfile/prisonerLocationHistory.js
jamesak-oncloud9: sure will revert it to how it was before

backend/api/prisonApi.js
jamesak-oncloud9: not used will remove

jamesak-oncloud9: not used will remove

backend/tests/prisonerLocationHistory.test.js
jamesak-oncloud9: more tests added

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -26,11 +56,24 @@ module.exports = ({ prisonApi, logError }) => async (req, res) => {
...
     const userCaseLoadIds = userCaseLoads.map(caseLoad => caseLoad.caseLoadId)
-
     const { bookingId, firstName, lastName } = prisonerDetails
-
     const currentPrisonerDetails = locationHistory.find(record => record.bookingId === bookingId) || {}
 
+    const { movementMadeBy, assignmentReason, bedAssignmentHistorySequence } = currentPrisonerDetails
+    const movementMadeByName = await fetchStaffName(res.locals, movementMadeBy, prisonApi)
+    const assignmentReasonName = assignmentReason
+      ? await fetchReasonDescription(res.locals, assignmentReason, caseNotesApi)
```
JamieDraperUK, line 65: Don't fix now but when for next time. You could simplify this by doing:
`
    const assignmentReasonName = assignmentReason && await fetchReasonDescription(res.locals, assignmentReason, caseNotesApi)
     
`

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -11,8 +11,38 @@ const {
...
-module.exports = ({ prisonApi, logError }) => async (req, res) => {
+const fetchStaffName = (context, staffId, prisonApi) =>
+  prisonApi.getStaffDetails(context, staffId).then(staff => formatName(staff.firstName, staff.lastName))
+
+const fetchReasonDescription = (context, assignmentReasonCode, caseNotesApi) =>
+  caseNotesApi
+    .getCaseNoteTypes(context)
+    .then(caseNoteTypes => caseNoteTypes.find(type => type.code === 'MOVED_CELL'))
+    .then(cellMoveTypes => {
+      return cellMoveTypes?.subCodes.map(subType => ({
```
JamieDraperUK, line 24: Could be inlined:
 .then(cellMoveTypes =>  cellMoveTypes?.subCodes.map(subType => ({
        value: subType.code,
        text: subType.description,
      }))
  )`

`

integration-tests/integration/prisonerProfile/prisonerQuickLook.spec.js
```diff
@@ -9,7 +9,7 @@ const quickLookFullDetails = {
   prisonerDetails: [
     {
       imprisonmentStatusDesc: 'Adult Imprisonment Without Option CJA03',
-      dateOfBirth: '1998-12-01',
+      dateOfBirth: '1999-12-01',
```
JamieDraperUK, line 12: Be better to set this date dynamically using moment.

JamieDraperUK (dismissed):
LGTM

integration-tests/integration/prisonerProfile/prisonerQuickLook.spec.js
jamesak-oncloud9: will be done in separate PR

JamieDraperUK (approved):

PR #1402, opened 2020-11-26
NN-3068 fix prisoner cell history tests

JamieDraperUK:
- This test failed locally due to expecting date times to be UTC


integration-tests/integration/prisonerProfile/prisonerCellHistory.spec.js
```diff
@@ -27,7 +27,7 @@ context('Prisoner cell history', () => {
             {
               agencyId: 'MDI',
               assignmentDate: '2020-05-01',
-              assignmentDateTime: '2020-05-01T12:48:33.375Z',
+              assignmentDateTime: '2020-05-01T12:48:33.375',
```
JamieDraperUK, line 30: Defining the dates times like this makes them UTC which will work on our builds but not locally.

jsrobertson (approved):
Hero

PR #1405, opened 2020-12-01
NN-3034: default on 404

jamesak-oncloud9:
Screenshot 2020-12-02 at 10 18 55


backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -72,7 +79,7 @@ module.exports = ({ prisonApi, whereaboutsApi, caseNotesApi, logError }) => asyn
       caseNotesApi,
       whereaboutsApi
     )
-    const whatHappenedDetailsToShow = isBlank(whatHappenedDetails) ? 'Not entered' : whatHappenedDetails
+    const whatHappenedTrimmedToSize = trimString(whatHappenedDetails, whatHappenedMaxLength)
```
JamieDraperUK, line 82: Why are you trimming the comment?

backend/controllers/prisonerProfile/prisonerLocationHistory.js
```diff
@@ -29,19 +31,26 @@ const fetchReasonDescription = (context, assignmentReasonCode, caseNotesApi) =>
...
-    .then(caseNote => caseNote.text)
-
+) => {
+  try {
+    return await whereaboutsApi
+      .getCellMoveReason(context, bookingId, bedAssignmentHistorySequence)
+      .then(cellMoveReason => caseNotesApi.getCaseNote(context, offenderNo, cellMoveReason.cellMoveReason.caseNoteId))
+      .then(caseNote => caseNote.text)
+  } catch (err) {
+    if (err?.response?.status === 404) {
```
JamieDraperUK, line 48: We generally prefer this style of if block:
`
if (err?.response?.status === 404)  return 'Not entered'
`

backend/controllers/prisonerProfile/prisonerLocationHistory.js
jamesak-oncloud9: due to comment in the ticket, but now removed

integration-tests/integration/prisonerProfile/prisonerLocationHistory.spec.js
```diff
@@ -268,91 +268,4 @@ context('Prisoner location history', () => {
         .contains('A long comment about what happened on the day to cause the move.')
     })
   })
-
-  context('Cell move comment functionality', () => {
```
JamieDraperUK, line 272: Why have you removed this integration test?

jamesak-oncloud9: I added this test before based on the wrong assumption that we could get an empty comment back, we only default on 404 and the system doesn't allow empty comment just like it doesn't allow over 4000 chars

JamieDraperUK: So should this integration test be tweaked to trigger the 404? That way the test is still valid.

jamesak-oncloud9: these tests already exist please see above lines 209 and 219

JamieDraperUK: Cant see anything in the integration file that relates to defaults. Looked at the following lines 209 and 219.

jamesak-oncloud9: the test that was removed asserts that if a rest call is made and the returned comment is of zero length then we default since this would not happen the test was removed. However, if this could happen? I'm happy to reinstate the test and make further code changes to support that behavior. The only time we default is when we get a 404 which are covered test on lines 209 and 219

JamieDraperUK (approved):
LGTM

PR #1408, opened 2020-12-02
DT-1078 Enable deleting secure casenotes

electrified:
Secure case notes and amendments can be deleted if the user has the role `DELETE_SENSITIVE_CASE_NOTES`. No other users will see this functionality.

Initially this will be given to admins, to save people on call from having to delete them via API calls which is easy to get wrong.

Screenshot from 2020-12-02 16-43-48
Screenshot from 2020-12-02 16-46-26
Screenshot from 2020-12-02 16-46-32


bell-pa (approved):
LGTM

PR #1426, opened 2020-12-09
NN-2833 New DPS homepage

jsrobertson:
image
*Not all card links are shown in screenshot


backend/config.js
```diff
@@ -29,6 +29,22 @@ module.exports = {
     port: process.env.REDIS_PORT || 6379,
     password: process.env.REDIS_PASSWORD,
   },
+  applications: {
```
jsrobertson, line 32: Created this because didn't feel right putting them within`apis`.  Think some of the others should be moved here too at a later date, `omic` and `bookVideoLink` for example.  Happy to discuss if you think otherwise though.

backend/controllers/homepage/homepage.js
```diff
@@ -0,0 +1,165 @@
+const {
+  applications: { licences, manageaccounts, moic, pecs },
+  apis: { omic, useOfForce, pathfinder, categorisation, soc },
+} = require('../../config')
+
+const getTasks = ({ activeCaseLoadId, locations, staffId, whereaboutsConfig }) => [
```
jsrobertson, line 6: Potential for improving this I think (calculating `enabled` along with the roles etc) but couldn't quite get it right so went with this method for now.

helm_deploy/values-dev.yaml
```diff
@@ -49,3 +49,5 @@ env:
   BVL_URL: https://book-video-link-dev.prison.service.justice.gov.uk
   MOIC_URL: https://allocation-manager-staging.apps.live-1.cloud-platform.service.justice.gov.uk
   OMIC_URL: https://dev.manage-key-workers.service.justice.gov.uk
+  MANAGE_AUTH_ACCOUNTS_URL: "https://manage-hmpps-auth-accounts-dev.prison.service.justice.gov.uk"
+  PECS_URL: "https://hmpps-book-secure-move-frontend-staging.apps.live-1.cloud-platform.service.justice.gov.uk"
```
JamieDraperUK, line 53: Don't need the double quotes for these ones.

backend/controllers/homepage/homepage.js
JamieDraperUK: Yeah I agree it would be nice to turn enabled into a function and let that decide, keeping all the logic in one place. Also they might be good candidates for class types. Happy to help with this in later PR's if you like?

JamieDraperUK (approved):
Nice work! Page looks amazing.

JamieDraperUK (approved):
LGTM

PR #1406, opened 2020-12-02
NN-3060 cell move filter update

jsrobertson:
Disregard cell attributes and have All, Single or Multiple options in cell type dropdown and filter based on the `cell.capacity` value.
image

PR #1429, opened 2020-12-11
NN-3091 Update cell move search options to filter on Single or Multiple occupancy

jsrobertson:
image

Very similar to #1406


JamieDraperUK (approved):
LGTM

PR #1432, opened 2020-12-14
NN-3095 map default route to the home page

JamieDraperUK:
At the moment when making a request to '/' the app would return the bundle.js and and static resources. However, we want to reserve that url for our home page. So this PR does the following:

- Move the transpiled react JS to /app
- Move all static resource to /static

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -12,9 +12,9 @@ image:
 ingress:
   enabled: true
   enable_whitelist: false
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -29,6 +29,14 @@ const plugins = [
       },
     ],
   }),
+  new CopyWebpackPlugin({
```
JamieDraperUK, line 32: Copy images over so that the shared header icon still loads.

PR #1435, opened 2020-12-14
Nn 3095

JamieDraperUK:
At the moment when making a request to '/' the app would return the bundle.js and and static resources. However, we want to reserve that url for our home page. So this PR does the following:

- Move the transpiled react JS to /app
- Move all static resource to /static

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -26,6 +26,9 @@ const setupApiRoutes = require('./setupApiRoutes')
 const setupReactRoutes = require('./setupReactRoutes')
 const phaseNameSetup = require('./phaseNameSetup')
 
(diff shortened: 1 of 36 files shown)


backend/index.js
```diff
@@ -67,9 +70,7 @@ app.use(
 
 app.use(setupReactRoutes())
 
-app.use((req, res) => {
-  res.redirect(config.app.notmEndpointUrl)
-})
+app.use(homepageController({ ...apis, logError }))
```
JamieDraperUK, line 73: Going to add in a 404 page handler in the next PR.

jsrobertson (approved):

PR #1447, opened 2020-12-16
NN-3057 Update move validation page to 'consider risks'.

jsrobertson:
localhost_3002_prisoner_G6123VU_cell-move_consider-risks_cellId=25718


backend/controllers/cellMove/considerRisks.js
```diff
@@ -0,0 +1,211 @@
+const moment = require('moment')
```
jsrobertson, line 1: Used to be `moveValidation.js`.

views/cellMove/considerRisks.njk
```diff
@@ -0,0 +1,209 @@
+{% extends "../partials/layout.njk" %}
```
jsrobertson, line 1: Used to be `moveValidation.njk`

JamieDraperUK (approved):
LGTM

PR #1449, opened 2020-12-17
NN-2992 Add cswap occupancy number to establishment roll page

jsrobertson:
Added **No cell allocated** value (cwap occupancy) and updated layout.

image


JamieDraperUK (approved):
LGTM!

PR #1450, opened 2020-12-18
NN-3073 move to a multi stage build and target buster slim

JamieDraperUK:


backend/index.js
```diff
@@ -20,17 +20,18 @@ const setupWebSecurity = require('./setupWebSecurity')
...
 const setupReactRoutes = require('./setupReactRoutes')
 const phaseNameSetup = require('./phaseNameSetup')
-
 const pageNotFound = require('./setUpPageNotFound')
 
 const { logError } = require('./logError')
 const homepageController = require('./controllers/homepage/homepage')
 
+// eslint-disable-next-line  global-require
+const setupWebpackForDev = !config.app.disableWebpack && require('./setupWebpackForDev')
```
JamieDraperUK, line 33: There are no dev dependencies in the docker image. This allows us to bypass any unnecessary module imports.

Dockerfile
```diff
@@ -19,16 +37,36 @@ RUN ln -snf "/usr/share/zoneinfo/$TZ" /etc/localtime && echo "$TZ" > /etc/timezo
...
+
+COPY --from=builder --chown=appuser:appgroup \
+        /app/src/dateHelpers.js ./src/dateHelpers.js
+
+COPY --from=builder --chown=appuser:appgroup \
+        /app/src/BulkAppointments/RecurringAppointments.js ./src/BulkAppointments/RecurringAppointments.js
+
 
 ENV PORT=3000
+ENV DISABLE_WEBPACK=true
```
JamieDraperUK, line 68: All dev dependencies have been removed due to running: `npm prune --production`

backend/index.js
```diff
@@ -48,6 +50,9 @@ app.use(setupAuth({ oauthApi: apis.oauthApi, tokenVerificationApi: apis.tokenVer
 app.use(setupWebpackForDev())
```
petergphillips, line 50: don't you need to remove this line too then?

backend/index.js
```diff
@@ -45,9 +47,11 @@ app.use(setupRedirects())
 app.use(setupStaticContent())
 app.use(setupWebSession())
 app.use(setupAuth({ oauthApi: apis.oauthApi, tokenVerificationApi: apis.tokenVerificationApi }))
-app.use(setupWebpackForDev())
 
 app.use(currentUser({ prisonApi: apis.prisonApi, oauthApi: apis.oauthApi }))
+
+if (!config.app.disableWebpack) app.use(setupWebpackForDev())
```
petergphillips, line 53: any reason for moving below the `currentUser` stuff - doesn't this mean that the webpack assets will get the current user in dev?

petergphillips (approved):
LGTM

backend/index.js
JamieDraperUK: No reason really. Technically a webpack module could (I think) make use of req.session, however this is just for dev so I'm not too worried.

PR #1454, opened 2020-12-21
DCS-759 Preventing middleware from being called twice

andrewrlee:
Removing unnecessary session data clean and fixing issue where middleware called twice. 

Two routes on the same path will share any middleware added by `router.use`

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -57,6 +57,13 @@ app.use(
   })
 )
 app.use(csrf())
+app.use((req, res, next) => {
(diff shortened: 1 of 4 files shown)


backend/index.js
```diff
@@ -54,6 +57,13 @@ app.use(
   })
 )
 app.use(csrf())
+app.use((req, res, next) => {
```
andrewrlee, line 60: extracted from `currentUser` this needs to happen after csrf registration

petergphillips (approved):
LGTM

PR #1460, opened 2021-01-05
NN-3122 bump packages

JamieDraperUK:


backend/api/oauthApi.js
```diff
@@ -24,8 +24,7 @@ const oauthApiFactory = (client, { clientId, clientSecret, url }) => {
   const userDetails = (context, username) => get(context, `/api/user/${username}`)
 
   const oauthAxios = axios.create({
-    baseURL: url,
-    url: 'oauth/token',
+    baseURL: `${url}/oauth/token`,
```
JamieDraperUK: A recent change in the way properties are copied over for post, put, and create method types, caused the requests to be sent with with an incomplete url. It would only make a request using the baseURL, ignoring url completely.  It seems url is no longer supported, this change was introduced in https://github.com/axios/axios/pull/2844/files

PR #1466, opened 2021-01-07
NN-3125 respond with correct http status codes

JamieDraperUK:
- This PR simply locates all try catch blocks and ensures the correct http status code is returned. 

_There is a ticket to move to global error handling. This is a quick fix to get insights to work correctly_


petergphillips (approved):
LGTM

PR #1467, opened 2021-01-08
NN-2909 New header

jsrobertson:
**Desktop:**
image

**Tablet:**
image

**Mobile:**
image


sass/application.scss
```diff
@@ -76,6 +76,11 @@ $govuk-page-width: 1170px;
   text-align: right;
 }
 
+.right-content {
```
jsrobertson, line 79: This was originally in header.scss so just moved it here for now.  Not a fan of this class but couldn't delete as it's used elsewhere at the moment.

backend/phaseNameSetup.js
```diff
@@ -4,6 +4,6 @@ const preprodText = 'This test version of DPS contains real data which may be up
 
 module.exports = (app, config) => {
```
jsrobertson, line 5: A lot of this feels unnecessary.  Will raise a ticket to redo.

views/partials/header.njk
```diff
@@ -1,53 +1,45 @@
...
+
+      <a class="dps-header__link dps-header__title__service-name" href="/">Digital prison services</a>
     </div>
+
+    <nav aria-label="Account navigation">
+      <ul class="dps-header__navigation">
+        <li class="dps-header__navigation__item">
+          <a class="dps-header__link" href="{{ authUrl + 'account-details' }}">
+            {{ user.displayName }}
+            <span class="dps-header__link__sub-text">Manage your details</span>
```
petergphillips, line 22: Why is this _Manage your details_ instead of _Manage account details_ in auth?  Should we change the auth one to match?

views/partials/header.njk
```diff
@@ -1,53 +1,45 @@
...
+  <span class="location-bar__location">
+    {% if phaseName %}
+      {{ govukTag({
+        text: phaseName,
+        classes: 'govuk-!-margin-right-1 ' + phaseNameColour
+      }) }}
+    {% endif %}
+    {{ user.activeCaseLoad.description }}
+  </span>
+  <a class="location-bar__link" href="/change-caseload">Change location</a>
```
petergphillips, line 44: will this always now be displayed?  I thought there was some logic around only showing if they can actually change caseload?

petergphillips: i.e. `if user.allCaseloads.length > 1 %}`

views/partials/header.njk
```diff
@@ -1,53 +1,45 @@
...
-                    </label>
-                </div>
-            </div>
-        {% endif %}
+        <image src="/assets/images/moj-logotype-crest.png" xlink:href="" class="moj-header__logotype-crest-fallback-image" width="40" height="40"></image>
+      </svg>
+
+      <a class="dps-header__link dps-header__title__organisation-name" href="/">HMPPS</a>
+
+      <a class="dps-header__link dps-header__title__service-name" href="/">Digital prison services</a>
```
petergphillips, line 14: shouldn't this be capitalised i.e. _Digital Prison Services_?

views/partials/header.njk
```diff
@@ -1,53 +1,45 @@
...
-    </div>
+<div class="location-bar">
+  <span class="location-bar__location">
+    {% if phaseName %}
+      {{ govukTag({
+        text: phaseName,
+        classes: 'govuk-!-margin-right-1 ' + phaseNameColour
+      }) }}
+    {% endif %}
+    {{ user.activeCaseLoad.description }}
```
petergphillips, line 42: think `user.activeCaseLoad` can be null if not a NOMIS user

petergphillips: but maybe not looking at the old code...

views/partials/header.njk
jsrobertson: Just double checked with Eddie, will change to _Manage account details_

views/partials/header.njk
jsrobertson: Well spotted, will update.

petergphillips (approved):
LGTM

PR #1480, opened 2021-01-19
NN-3046 Conditionally show damage obligations tab/link on private cash page

jsrobertson:


backend/controllers/prisonerProfile/prisonerFinances/prisonerDamageObligations.js
```diff
@@ -58,6 +58,7 @@ module.exports = ({ prisonApi }) => async (req, res) => {
         offenderNo,
       },
       rows,
+      showDamageObligationsLink: true,
       totalOwed: formatCurrency(totalOwed),
     })
   } catch (error) {
```
JamieDraperUK, line 64: You should be able to use the new error handler. If the redirect url is the original one you can just let the error bubble up. If not you need to set res.locals.redcirectUrl and then re-throw the error.

jsrobertson: Nice one, will have a look

JamieDraperUK (dismissed):

JamieDraperUK (approved):
LGTM

PR #1492, opened 2021-02-01
NN-3100 No cell allocated page

jsrobertson:
image

### DIFF
--- a/backend/controllers/establishmentRoll/noCellAllocated.js
+++ b/backend/controllers/establishmentRoll/noCellAllocated.js
@@ -11,11 +11,21 @@ module.exports = ({ prisonApi }) => async (req, res) => {
 
     const offenderNumbers = offendersInCellSwap.map(offender => offender.offenderNo)
 
-    const offendersWithNoCell = await Promise.all(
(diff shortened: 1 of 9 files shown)


backend/controllers/establishmentRoll/noCellAllocated.js
```diff
@@ -0,0 +1,49 @@
+const { formatName, putLastNameFirst, getTime, stripAgencyPrefix } = require('../../utils')
```
JamieDraperUK, line 1: It might be more efficient to make one call to:  `POST /api/prisoners `with ` {"offenderNos": [
    "string"
  ]}`

JamieDraperUK: Same for staff details. Use `POST /api/users/list` with ```
[
  "string"
]
```

jsrobertson: Good shout, forgot about that endpoint. Ta

JamieDraperUK (approved):
LGTM

PR #1493, opened 2021-02-02
NN-3100 Only show Allocate cell links if user has cell move role

jsrobertson:
**Without role:**
image

**With role:**
image


JamieDraperUK (approved):
LGTM

PR #1495, opened 2021-02-04
DCS-842 Removing references to VCCs in PSH

andrewrlee:
We never send emails to VCC admins in PSH so this stuff can go


JamieDraperUK (approved):
:1st_place_medal:

PR #1496, opened 2021-02-08
NN-3134 Switch font weight on prisoner listed values

jsrobertson:
Also add Prisoners listed value to individual attendance reason statistics page.


JamieDraperUK (approved):
LGTM

PR #1498, opened 2021-02-08
NN-3145 filter cleared holdings out of the list

JamieDraperUK:


views/prisonerProfile/prisonerFinance/privateCash.njk
```diff
@@ -33,8 +33,7 @@
     {{ govukTable({
       head: [
         { text: "Date" },
-        { text: "Money in" },
-        { text: "Money out" },
+        { text: "Amount" },
```
petergphillips, line 36: Isn't this reliant on `mergeAmounts` being `true`?

backend/shared/createTransactionViewModel.js
```diff
@@ -1,14 +1,18 @@
...
   transactions.map(transaction => {
     const { description: prisonName } = prisons.find(agency => transaction.agencyId === agency.agencyId)
     const { penceAmount, currentBalance } = transaction
 
     return [
       { text: transaction.entryDate && formatTimestampToDate(transaction.entryDate) },
-      { text: transaction.postingType === 'CR' ? formatCurrency(penceAmount / 100) : '' },
-      { text: transaction.postingType === 'DR' ? formatCurrency(penceAmount / 100) : '' },
+      ...(mergeAmounts
+        ? [{ text: (penceAmount && formatCurrency(penceAmount / 100)) || '' }]
```
petergphillips, line 11: I would have thought that `formatCurrency` should be able to cope with `penceAmount` being `0` or `false`?  Is it that if the value is `0` you don't want to display anything

petergphillips (approved):
LGTM

views/prisonerProfile/prisonerFinance/privateCash.njk
JamieDraperUK: It's always true for that section. Having a central mapper that works for difference screens is a bit of a smell here and is causing confusion.

backend/shared/createTransactionViewModel.js
JamieDraperUK: Agreed. Will improve it in the next PR.

PR #1518, opened 2021-02-17
NN-3144: Add home link to cell move confirmation page

sp-watson:


sp-watson:
Doing it again, as there is a much easier way

PR #1529, opened 2021-02-19
NN-3182 notification banner

JamieDraperUK:
image


backend/services/notificationCookie.js
```diff
@@ -0,0 +1,14 @@
+const COOKIE_NAME = 'ui-notification-bar'
+const notificationKey = ({ id, revision }) => `${id}-${revision}`
+
+const markAsDismissed = (res, { id, revision }) => {
+  const oneYear = 52 * 24 * 3600000
+  return res.cookie(COOKIE_NAME, notificationKey({ id, revision }), { maxAge: oneYear, httpOnly: true })
```
andrewrlee, line 6: [We erred on caution and signed and set the secure flag in production ](https://github.com/ministryofjustice/hmpps-book-video-link/blob/main/backend/routes/amendBooking/state.ts)

Wasn't sure if this was the sort of thing that might appear on a pen test or not. 

obviously you are not storing any critical data in here

JamieDraperUK: I know what you're saying but in this case I think it's fine as it only drives the viability of the notification bar.

backend/middleware/notificationHandler.js
```diff
@@ -0,0 +1,47 @@
...
+
+const options = {
+  renderMark: {
+    [MARKS.BOLD]: text => `<p class="govuk-!-font-weight-bold">${text}</p>`,
+    [MARKS.ITALIC]: text => `<em>${text}</em>`,
+  },
+  renderNode: {
+    [INLINES.HYPERLINK]: (node, next) => {
+      // Update to @govuk-react/link when out of Alpha
+      return `<a class="link" href="${node.data.uri}">
```
jsrobertson, line 13: class="govuk-link" and can get rid of the reference to govuk-react/link

backend/middleware/notificationHandler.js
```diff
@@ -0,0 +1,47 @@
+const moment = require('moment')
+const { INLINES, MARKS } = require('@contentful/rich-text-types')
+const { documentToHtmlString } = require('@contentful/rich-text-html-renderer')
+
+const options = {
+  renderMark: {
+    [MARKS.BOLD]: text => `<p class="govuk-!-font-weight-bold">${text}</p>`,
```
jsrobertson, line 7: <strong>${text}</strong> would be better here I reckon.  We do the same in content.js

jsrobertson (approved):
LGTM

PR #1532, opened 2021-02-22
NN-3008 CSRA history page

jsrobertson:
Screenshot 2021-02-22 at 12 11 11


backend/controllers/prisonerProfile/prisonerCsraHistory.js
```diff
@@ -0,0 +1,58 @@
...
+
+const csraOptions = [
+  { value: 'STANDARD', text: 'Standard' },
+  { value: 'HI', text: 'High' },
+  { value: 'PEND', text: 'Pending' },
+]
+
+module.exports = ({ prisonApi }) => async (req, res) => {
+  const { offenderNo } = req.params
+  const { csra, location } = req.query
```
JamieDraperUK, line 11: We need to start handling missing query parameters like these. Need to either redirect or respond with a something went wrong page and a HTTP 400 status.

backend/tests/prisonerCsraHistory.test.js
```diff
@@ -0,0 +1,202 @@
+const prisonerCsraHistory = require('../controllers/prisonerProfile/prisonerCsraHistory')
```
JamieDraperUK, line 1: Is there a test that covers sorting?

JamieDraperUK (approved):
LGTM

backend/tests/prisonerCsraHistory.test.js
jsrobertson: Yeah although there's no specific labelled test for it, it's covered under the `should render the correct template with the correct values` test. The mock data is ordered as **2011-06-02, 2014-11-10, 2013-11-10** but gets passed to the template in the correct descending order.

PR #1535, opened 2021-02-22
NN-3186 fix stats page bug

JamieDraperUK:
There aren't any int tests for this page so I'll add some in the next PR


petergphillips (approved):
LGTM

PR #1537, opened 2021-02-23
NN-3115 Update CSRA info in prisoner profile header

jsrobertson:
Combine CSRA and last review date in prisoner profile header and add link to new CSRA history page:
image


JamieDraperUK (approved):
LGTM

PR #1538, opened 2021-02-23
NN-3197 handle timeouts

JamieDraperUK:
The select activity location page uses a very slow query that generally times out on the first few tries. This ticket stops these types of errors polluting our azure logs. 

Longer time we need to investigate if elastic search is a viable solution to this problem.


jsrobertson (approved):
Nice one

PR #1539, opened 2021-02-23
Nn 3178 handle validation errors for case notes

sp-watson:
Handle prison-api validation of case notes correctly


backend/tests/caseNote.test.js
```diff
@@ -187,6 +188,38 @@ describe('case note management', () => {
...
+          body: {
+            type: 'P',
+            subType: 'PI',
+            date: '20/07/2020',
+            hours: '10',
+            minutes: '10',
+            text: 'test',
+          },
+        }
+        const error400 = {
```
JamieDraperUK, line 204: Not a massive issue but you've tested this slightly differently above.

backend/controllers/caseNote.js
```diff
@@ -178,6 +178,30 @@ const caseNoteFactory = (prisonApi, caseNotesApi) => {
       // has been handled by the above validations
     }
 
+    if (errors.length === 0) {
```
JamieDraperUK, line 181: Might be neater pulling this out into a function

JamieDraperUK (approved):
LGTM

backend/controllers/caseNote.js
sp-watson: I tried that, and it didn't improve readability for me due to the way it returned the error(s) or threw one.

Maybe you had a specific impl in mind that might be better?

backend/tests/caseNote.test.js
sp-watson: Fixed it. I did that originally and moved so I could change quicker as things weren't working

JamieDraperUK (approved):
LGTM :+1:

PR #1559, opened 2021-03-02
NN-3131 webpack 5 upgrade

jsrobertson:


static/styles/gov-uk.css
```diff
@@ -1,4 +1,4 @@
-.font-large,
+/* .font-large,
```
JamieDraperUK, line 1: Did you mean to check this in?

JamieDraperUK (approved):

JamieDraperUK (approved):

PR #1572, opened 2021-03-10
NN-3114 CSRA details page

jsrobertson:
CSRA details page:
image

Link from CSRA history page:
image

### DIFF
--- a/backend/controllers/prisonerProfile/prisonerCsraReview.js
+++ b/backend/controllers/prisonerProfile/prisonerCsraReview.js
@@ -1,13 +1,11 @@
 const { putLastNameFirst, formatName, formatTimestampToDate } = require('../../utils')
+const { notEnteredMessage } = require('../../common-messages')
 
-const csraOptions = [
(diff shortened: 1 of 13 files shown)


backend/controllers/prisonerProfile/prisonerCsraReview.js
```diff
@@ -0,0 +1,78 @@
...
+  HI: 'High',
+}
+
+module.exports = ({ prisonApi }) => async (req, res, next) => {
+  const { offenderNo } = req.params
+  const { assessmentSeq } = req.query
+
+  try {
+    const prisonerDetails = await prisonApi.getDetails(res.locals, offenderNo)
+    const { bookingId, firstName, lastName } = prisonerDetails
```
sp-watson, line 16: I don't think this works for reviews from a previous location. I believe the bookingId needs to come from the link - it is part of the details returned for each CSRA in the history page. BTW you can test with prisoner `G2458GG` on T3

sp-watson (approved):

PR #1575, opened 2021-03-11
NN-3219 Add Google Tag Manager and active case load selector

jsrobertson:
Matching selector configured in GTM dashboard:
Screenshot 2021-03-10 at 16 03 32

We still need the Google analytics ID for now, for the React App and the server side raiseAnalyticsEvent.


JamieDraperUK (approved):
LGTM!

PR #1576, opened 2021-03-11
NN-3114 Update csra reviewed by to add assessment committee

jsrobertson:
Also updated breadcrumbs with page title and link back to CSRA history


backend/controllers/prisonerProfile/prisonerCsraReview.js
```diff
@@ -61,7 +70,7 @@ module.exports = ({ prisonApi }) => async (req, res, next) => {
         },
         {
           key: { text: 'Reviewed by' },
-          value: { text: formatName(staffDetails.firstName, staffDetails.lastName) },
+          value: { text: originalClassificationCode ? `${assessorAuthority} - ${assessorName}` : assessorName },
```
sp-watson, line 73: i don't think we need the `originalClassificationCode` check do we?

sp-watson (approved):

PR #1581, opened 2021-03-12
Nn 3220 use new csra classification prop

sp-watson:
Use the csraClassificationCode added to the prison-api endpoint for full offender details, instead of the csra prop, which just gave the last result

### DIFF
--- a/backend/shared/csraHelpers.js
+++ b/backend/shared/csraHelpers.js
@@ -1,0 +1,10 @@
+const csraTranslations = {
+  LOW: 'Low',
+  STANDARD: 'Standard',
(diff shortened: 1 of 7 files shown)


backend/shared/csraHelpers.js
```diff
@@ -0,0 +1,9 @@
+const csraTranslations = {
+  LOW: 'Low',
+  STANDARD: 'Standard',
+  HI: 'High',
```
JamieDraperUK, line 4: :grin: - Should be HIGH?

JamieDraperUK (approved):
LGTM

PR #1586, opened 2021-03-15
NN-3175 Update back/cancel link on confirm cell move page

jsrobertson:
Link text changed to cancel as there is now multiple entry points to this page. Returns to search for cell page by default, but should use the referring url, if available. 

Screenshot 2021-03-15 at 11 07 41


JamieDraperUK (approved):
LGTM

PR #1587, opened 2021-03-15
Revert "Nn 3220 use new csra classification prop"

sp-watson:
Reverts ministryofjustice/prisonstaffhub#1581


sp-watson:
Not needed as we are going to release it

PR #1593, opened 2021-03-17
NN-3187 cell move handle cswap recp

sp-watson:
On the cell move journey, do not show the acronyms of the non-cell locations, e.g. CSWAP or RECP - show full descriptions.

Also, do not show links to the location history of non-cells.


JamieDraperUK (approved):
LGTM

PR #1594, opened 2021-03-17
NN-3149 Update content on cell move consider risks page

jsrobertson:
When there are no inmates in the location you are moving a prisoner too (usually single occupancy):
- Do not display CSRA info
- Confirmation label should not mention any prisoner names and just say, **Are you sure you want to select this cell?**

image


JamieDraperUK (approved):
LGTM

sp-watson (approved):

PR #1595, opened 2021-03-18
NN-3187: Show description if prisoner in temporary location

sp-watson:
The prisoner details page still shows CSWAP - it needs to show "No cell allocated"


JamieDraperUK (approved):
LGTM

PR #1608, opened 2021-03-26
NN-3184: Handle concurrent closing of alerts

sp-watson:


backend/controllers/alert.js
```diff
@@ -100,6 +100,14 @@ const alertFactory = (oauthApi, prisonApi, referenceCodesService) => {
         oauthApi.userRoles(res.locals),
       ])
 
+      if (alert && alert.expired) {
```
jsrobertson, line 103: Can do `alert?.expired`

sp-watson: Yeah sounds sensible. I only moved that code, but i will fix

jsrobertson (approved):

PR #1620, opened 2021-04-08
NN-3204: Handle confirm appointment refreshes

sp-watson:
Do not show the "Sorry" page when a user refreshes the confirm-appointment page, instead redirect back to the start of the journey


backend/controllers/appointments/confirmAppointment.js
```diff
@@ -22,7 +22,10 @@ const confirmAppointmentFactory = ({ prisonApi, appointmentsService, logError })
       )
 
       const appointmentDetails = req.flash('appointmentDetails')
-      if (!appointmentDetails?.length) throw new Error('Appointment details are missing')
+      if (!appointmentDetails?.length) {
+        res.redirect(prisonUser ? `/prisoner/${offenderNo}` : '/videolink/prisoner-search')
```
JamieDraperUK, line 26: Be better to do a return res.redirect

sp-watson: Ha - that's what i did originally but eslint complained that it was an inconsistent return type (there is no return value otherwise) - https://app.circleci.com/pipelines/github/ministryofjustice/prisonstaffhub/3861/workflows/4382c341-35af-4ca4-ab78-4da536879af5/jobs/17620.

JamieDraperUK (approved):

PR #1626, opened 2021-04-20
NN-3263 Add new cell move homepage

jsrobertson:
image

Also removed a call to oAuth on the existing whereabouts homepage as none of the tasks are role dependant.

### DIFF
--- a/backend/controllers/whereabouts/whereaboutsHomepage.js
+++ b/backend/controllers/whereabouts/whereaboutsHomepage.js
@@ -41,21 +41,7 @@ const whereaboutsTasks = [
   },
 ]
 
-module.exports = oauthApi => async (req, res) => {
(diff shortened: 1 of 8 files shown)


backend/controllers/whereabouts/whereaboutsHomepage.js
```diff
@@ -41,21 +41,7 @@ const whereaboutsTasks = [
   },
 ]
 
-module.exports = oauthApi => async (req, res) => {
-  try {
-    const userRoles = await oauthApi.userRoles(res.locals)
```
jsrobertson, line 46: None of the tasks on whereabouts homepage are role dependant so a call to oAuth is not needed.

sp-watson (approved):

PR #1629, opened 2021-04-21
NN-3265: Cell move prisoner search page

sp-watson:


backend/controllers/cellMove/cellMovePrisonerSearch.js
```diff
@@ -0,0 +1,65 @@
...
+
+module.exports = ({ prisonApi }) => async (req, res) => {
+  const {
+    user: { activeCaseLoad },
+  } = res.locals
+  const { keywords } = req.query
+
+  if (!keywords) {
+    const errors =
+      keywords === undefined
```
JamieDraperUK, line 12: I'm not sure this check is needed. The if condition should be enough here.

backend/controllers/cellMove/cellMovePrisonerSearch.js
```diff
@@ -0,0 +1,65 @@
...
+
+  const context = {
+    ...res.locals,
+    requestHeaders: {
+      'Sort-Fields': 'lastName,firstName',
+      'Sort-Order': 'ASC',
+    },
+  }
+
+  const [prisoners] = await Promise.all([
```
JamieDraperUK, line 36: We only use Promise.all when making multiple requests to APIs, as it makes them concurrently. Really all you need here is: `const prisoners = await  prisonApi.getInmates(context, currentUserCaseLoad, {
      keywords,
      returnAlerts: 'true',
    })`

backend/controllers/cellMove/cellMovePrisonerSearch.js
```diff
@@ -0,0 +1,65 @@
...
+  const results =
+    prisoners &&
+    prisoners.map(prisoner => ({
+      ...prisoner,
+      assignedLivingUnitDesc: formatLocation(prisoner.assignedLivingUnitDesc),
+      name: putLastNameFirst(prisoner.firstName, prisoner.lastName),
+      alerts: alertFlagLabels.filter(alertFlag =>
+        alertFlag.alertCodes.some(
+          alert =>
+            prisoner.alertsDetails && prisoner.alertsDetails.includes(alert) && cellMoveAlertCodes.includes(alert)
```
JamieDraperUK, line 52: You might be able to take advantage of optionals here. So instead of 
`prisoner.alertsDetails && prisoner.alertsDetails.includes(alert)`
Could you do
`prisoner.alertsDetails?.includes(alert)`

integration-tests/integration/cellMove/cellMovePrisonerSearch.spec.js
```diff
@@ -0,0 +1,108 @@
...
+
+      cy.get('[data-test="prisoner-search-results-table"]').then($table => {
+        cy.get($table)
+          .find('tr')
+          .then($tableRows => {
+            cy.get($tableRows)
+              .its('length')
+              .should('eq', 3) // 2 results plus table header
+            expect($tableRows.get(1).innerText).to.contain(
+              '\tSmith, John\tA1234BC\tUNIT-1\t\nARSONIST\n\n\tView cell history\tChange cell'
```
JamieDraperUK, line 74: We found that asserting the rows like this becomes a maintenance problem. We've been mapping to an object the asserting on that. See prisonersUnaccountedFor.spec.js

views/cellMove/cellMovePrisonerSearch.njk
```diff
@@ -0,0 +1,131 @@
...
+  </h1>
+
+  <form novalidate="novalidate" class="prisoner-search__form" data-test="prisoner-search-form">
+    <div class="horizontal-form govuk-!-margin-bottom-3">
+      {{ govukInput({
+        id: "keywords",
+        name: "keywords",
+        value: formValues.keywords,
+        label: {
+            text: "Enter a prisoner&#8217;s name or number"
```
JamieDraperUK, line 50: This needs to be html otherwise it'll just show **&#8217;** And Andy will kill us :)

views/cellMove/cellMovePrisonerSearch.njk
```diff
@@ -0,0 +1,131 @@
...
+    </div>
+  </form>
+
+  {% if hasSearched %}
+    <p class="govuk-body pull-right"><strong>Prisoners listed:</strong> {{totalOffenders}}</p>
+
+    {% if results.length %}
+      {% set rows = [] %}
+      {% for prisoner in results %}
+        {% set prisonerImage = '<img src="/app/images/' + prisoner.offenderNo + '/data" alt="Image of ' + prisoner.name + '" class="prisoner-search__results-list__image" />' %}
```
JamieDraperUK, line 72: This doesn't look right. Are there other places we do this?

backend/controllers/cellMove/cellMovePrisonerSearch.js
sp-watson: That is what I started with, but I needed to distinguish between the first visit to the page (no keyword param in the URL) and a search without keywords (keyword param without a value). i couldn't see a better way of doing it then this - but there might be a construct that would do it better?!

backend/controllers/cellMove/cellMovePrisonerSearch.js
sp-watson: Ah yeah

backend/controllers/cellMove/cellMovePrisonerSearch.js
sp-watson: Yes, copied code!

integration-tests/integration/cellMove/cellMovePrisonerSearch.spec.js
sp-watson: ok

views/cellMove/cellMovePrisonerSearch.njk
sp-watson: oh

views/cellMove/cellMovePrisonerSearch.njk
sp-watson: yeah - prisonerSearch.njk

backend/controllers/cellMove/cellMovePrisonerSearch.js
JamieDraperUK: Oh I'm with you. I wonder if we should add a comment explaining the reasoning behind it?

backend/controllers/cellMove/cellMovePrisonerSearch.js
```diff
@@ -0,0 +1,58 @@
...
+const { putLastNameFirst, formatLocation } = require('../../utils')
+
+module.exports = ({ prisonApi }) => async (req, res) => {
+  const {
+    user: { activeCaseLoad },
+  } = res.locals
+  const { keywords } = req.query
+
+  if (!keywords) {
+    const firstCall = keywords === undefined
```
JamieDraperUK, line 11: You've already got the perfect name for this, _hasSearched_

sp-watson: hmm.. hasSearched currently means actually sent a search to the server - but I prob should change it

JamieDraperUK (approved):

PR #1630, opened 2021-04-21
NN-3330 Turn on complexity functionality for select prisons

JamieDraperUK:


backend/services/prisonerProfileService.js
```diff
@@ -142,9 +144,11 @@ module.exports = ({
 
     const canViewSocLink = Boolean(isSocUser && socDetails)
 
-    const complexityLevel = complexity.enabled && (await complexityApi.getComplexOffenders(context, [offenderNo]))
+    const complexityLevel =
+      isComplexityEnabledFor(agencyId) && (await complexityApi.getComplexOffenders(context, [offenderNo]))
+
     const isHighComplexity =
-      complexity.enabled && Boolean(complexityLevel?.length > 0 && complexityLevel[0]?.level === 'high')
+      isComplexityEnabledFor(agencyId) && Boolean(complexityLevel?.length > 0 && complexityLevel[0]?.level === 'high')
```
sp-watson, line 151: i don't think you need the call to `isComplexityEnabledFor` here - the `complexityLevel` will not be set if complexity is not enabled for this agency

sp-watson (dismissed):
One possible change, but good otherwise

backend/services/prisonerProfileService.js
JamieDraperUK: Ah, good point.

sp-watson (approved):

PR #1637, opened 2021-04-27
NN-3266: New cell move view residential location page

sp-watson:
image

### DIFF
--- a/backend/controllers/cellMove/cellMoveViewResidentialLocation.js
+++ b/backend/controllers/cellMove/cellMoveViewResidentialLocation.js
@@ -7,7 +7,7 @@ const prionsApiLocationId = async (res, whereaboutsApi, locationKey, userCaseLoa
     const locationIdWithSuffix = fullLocationPrefix?.locationPrefix
     return locationIdWithSuffix?.length < 1 ? '' : locationIdWithSuffix?.slice(0, -1)
   }
-  return userCaseLoad + '-' + locationKey
+  return `${userCaseLoad}-${locationKey}`
 }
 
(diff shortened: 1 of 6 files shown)


backend/controllers/cellMove/cellMoveViewResidentialLocation.js
```diff
@@ -0,0 +1,84 @@
+const { alertFlagLabels, cellMoveAlertCodes } = require('../../shared/alertFlagValues')
+const { putLastNameFirst, formatLocation } = require('../../utils')
+
+const prisonApiLocationId = async (res, whereaboutsApi, locationKey, userCaseLoad) => {
```
sp-watson, line 4: I decided to do the location-description-building here as the `whereabouts-api` also returns sub-locations. For those sub-locations there is no obvious, reliable way of generating the location description.

views/cellMove/cellMoveViewResidentialLocation.njk
```diff
@@ -0,0 +1,138 @@
...
+              text: "Name",
+              attributes: {
+                "aria-sort": "ascending"
+              }
+            },
+            { text: "Prison number" },
+            {
+              text: "Location",
+              attributes: {
+                "aria-sort": "ascending"
```
jsrobertson, line 113: I've since realised that only one table header (usually the name one) needs `"aria-sort": "ascending"`.  All the others should be `"aria-sort": "none"`. This will display the correct up/down arrows then but still allow the column to be sortable.

sp-watson: Ah ok - makes sense. Fixing

jsrobertson (approved):

PR #1661, opened 2021-05-07
NN-3372 Fix current npm security audit issues

jsrobertson:
Added [1677](https://www.npmjs.com/advisories/1677) to allowlist however it is because of a dependancy of jest.  There's an issue raised on their github which I've subscribed to so will keep an eye on https://github.com/facebook/jest/issues/11383.

Also removed [961](https://www.npmjs.com/advisories/961) from allowlist as no longer an issue for us.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,6 +6,7 @@ docs
 bin
 static
 audit-ci.json
+package.json
 package-lock.json
 cypress.json
 cypress-reporter-config.json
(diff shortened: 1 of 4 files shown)


karim10:
Hi @jsrobertson ,
Sorry for commenting directly on your repository. I have a quick question: how do you allow some packages in the npm audit script because I have the same issue with hosted-git-info.
Best

PR #1670, opened 2021-05-17
NN-3348 Add res location filter to appointments screen

jsrobertson:
Also changed url and some UI updates to filter section:
image


backend/routes/appointments/viewAppointmentsRouter.js
```diff
@@ -2,20 +2,35 @@ const moment = require('moment')
 const { serviceUnavailableMessage } = require('../../common-messages')
 const { getTime, properCaseName, formatName, getCurrentPeriod } = require('../../utils')
 
+const prisonApiLocationDescription = async (res, whereaboutsApi, locationKey, userCaseLoad) => {
```
sp-watson, line 5: Just reminded me - we ought to make this shared.. but not urgent

sp-watson (approved):

PR #1684, opened 2021-05-26
NN-3350: Confirm delete appointments page

sp-watson:


backend/tests/appointmentConfirmDeletion.test.js
```diff
@@ -0,0 +1,191 @@
+const appointmentConfirmDeletion = require('../controllers/appointmentConfirmDeletion')
```
sp-watson, line 1: Doing int tests later - possibly as part of full appt deletion journey

backend/controllers/appointmentConfirmDeletion.js
```diff
@@ -0,0 +1,86 @@
+const packAppointmentDetails = (req, details) => {
+  req.flash('appointmentDetails', details)
+}
+
+const unpackAppointmentDetails = req => {
+  const appointmentDetails = req.flash('appointmentDetails')
+  if (!appointmentDetails || !appointmentDetails.length) throw new Error('Appointment details are missing')
```
JamieDraperUK, line 7: This would result in HTTP 500 error being posted to dps_alerts. I'm not sure if this is the same but we've had situations previously where a user has hit F5 and triggered an error. Would it be better to redirect back to the start of the journey?

backend/controllers/appointmentConfirmDeletion.js
```diff
@@ -0,0 +1,86 @@
...
+      isRecurring,
+      additionalDetails,
+      basicDetails,
+      prepostData,
+      recurringDetails,
+      timeDetails,
+    } = appointmentDetails
+
+    // Save for possible re-showing on error
+    packAppointmentDetails(req, appointmentDetails)
```
JamieDraperUK, line 36: Not fully sure if this is needed. Wouldn't you just create new errors within the post request then redirect back to here?

backend/controllers/appointmentConfirmDeletion.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  const post = async (req, res) => {
+    const { confirmation, isRecurring, appointmentEventId } = req.body
+
+    if (!confirmation) {
+      const errors = []
+      errors.push({ text: 'Select yes if you want to delete this appointment', href: '#confirmation' })
+      return renderTemplate(req, res, errors)
```
JamieDraperUK, line 65: We should be doing a redirect back to the GET route here. The pattern is called PRG (Post, redirect, get)

backend/controllers/appointmentDetails.js
```diff
@@ -43,34 +43,50 @@ module.exports = ({ prisonApi, whereaboutsApi }) => async (req, res) => {
     prepostData['post-court hearing briefing'] = createLocationAndTimeString(videoLinkBooking.post)
   }
 
+  const additionalDetails = {
+    ...(videoLinkBooking && { courtLocation: videoLinkBooking.main.court }),
+    comments: appointment.comment || 'Not entered',
+    // addedBy: formatName(staffDetails.firstName, staffDetails.lastName),
```
JamieDraperUK, line 49: This is okay to go back in now we fall back to prison api, maybe one for later though.

backend/controllers/appointmentConfirmDeletion.js
sp-watson: I do - but the flash is reset just above this, so without this there would be an error but no data in the flash

backend/controllers/appointmentConfirmDeletion.js
```diff
@@ -0,0 +1,86 @@
+const packAppointmentDetails = (req, details) => {
+  req.flash('appointmentDetails', details)
```
JamieDraperUK, line 2: I might be missing something but why do we need to store the details? Don't we just need the appointment id?

backend/controllers/appointmentConfirmDeletion.js
sp-watson: Hmm.. I copied what was done elsewhere (e.g. bulkAppointmentsConfirm).

I also don't know why it is necessary?!

backend/controllers/appointmentDetails.js
sp-watson: John did that - I moved the code. Adding this back in is a bit more involved now we have 2 njks displaying this data, but we will do that next

backend/controllers/appointmentConfirmDeletion.js
sp-watson: The point is that the appointment details page makes lots of calls and does lots of calculations - so why do it all again here? Also, this is just a confirmation page - it shouldn't be accessible without going to the initial details page first, hence we don't even have the id in the route.

JamieDraperUK: It's to keep the code simpler. It's acceptable to make a request each time for the appointment details. It's a standard flow when it comes to web.

backend/controllers/appointmentConfirmDeletion.js
sp-watson: That was NN-3204 - actually we never understood why that was happening, just that it did. But thinking about it, yes, redirecting is probably better - it would have to be the `view-all-appointments` page though

backend/controllers/appointmentConfirmDeletion.js
JamieDraperUK: It's standard practice to stop the form from being resubmitted on page refresh.

sp-watson: Standard practice that we don't do everywhere?!

JamieDraperUK: I may of missed something. Can we have a call as it feels odd doing it all on here.

backend/controllers/appointmentConfirmDeletion.js
sp-watson: I did this because of the reasons given above and - more importantly - it has already been done for appointments (see the add appointment journey).

Personally I wouldn't want to change it for 2 reasons (if i understand what you are proposing correctly): 

- Code duplication is not good for future maintenance, especially when it is exactly the same code, as this would be w.r.t. the appointmentDetails page - but we could get round this by using a common component/class. Or would that make it too complicated?
- Inefficiency. Why get the same data twice?

It makes it difficult to know what the right thing to do is - there is no documentation for these standard practices AND code already exists that doesn't use these standard practices.

JamieDraperUK (approved):
LGTM :+1:

backend/controllers/appointmentConfirmDeletion.js
sp-watson: We will make a service to provide the appointment info, and change appt details and this page to use that. The flash will be removed.

backend/controllers/appointmentConfirmDeletion.js
sp-watson: We will change this, as showing info in a post request leads to problems if the user refreshes the page

PR #1686, opened 2021-05-26
NN-3351: New appointment deleted page

sp-watson:


backend/tests/appointmentDeleted.test.js
```diff
@@ -0,0 +1,41 @@
+const appointmentDeleted = require('../controllers/appointmentDeleted')
```
sp-watson, line 1: Going to write int tests along the confirm deletion int tests

JamieDraperUK (approved):
LGTM

PR #1687, opened 2021-05-27
NN-3424 Fixing failing test

andrewrlee:


backend/tests/addAppointment.test.js
```diff
@@ -13,6 +13,16 @@ describe('Add appointment', () => {
   const offenderNo = 'ABC123'
   const bookingId = 123
 
+  const twoDaysHence = moment().add(2, 'day')
+
```
bell-pa, line 17: Oooh! I like name of the reference. I shall make use of it henceforth.

bell-pa (approved):
LGTM

PR #1690, opened 2021-05-27
DCS-983 Send court ids when creating a BVL booking

andrewrlee:
We now send will send court ids when creating new bookings when we can tie this to a specific known court

We still will need to send the court name through if a user selects an 'other' court. These other courts won't appear in the BVL booking list

### DIFF
--- a/backend/controllers/appointments/prepostAppoinments.js
+++ b/backend/controllers/appointments/prepostAppoinments.js
@@ -250,7 +250,7 @@ const prepostAppointmentsFactory = ({
       preAppointmentDuration,
       preAppointmentLocation,
       postAppointmentLocation,
-      court,
+      court: courtId,
       otherCourt,
       otherCourtForm,
     } = req.body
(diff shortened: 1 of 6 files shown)


backend/controllers/appointments/prepostAppoinments.js
```diff
@@ -80,11 +80,8 @@ const prepostAppointmentsFactory = ({
   }
 
   const getCourts = async locals => {
-    const { courtLocations } = await whereaboutsApi.getCourtLocations(locals)
-    const formattedLocations = courtLocations.sort().reduce((courtList, court) => {
-      const key = court.replace(/\W+/g, '-').toLowerCase()
-      return { ...courtList, [key]: court }
-    }, {})
+    const items = await whereaboutsApi.getCourtLocations(locals)
+    const formattedLocations = Object.fromEntries(items.map(({ id, name }) => [id, name]))
```
bell-pa, line 84: That's better!

bell-pa (approved):
LGTM

PR #1691, opened 2021-05-27
NN-3350: Extract common appointment deletion view model to a service

sp-watson:
So the appointmentDetails and appointmentConfirmDeletion pages use the same logic.


backend/services/appointmentDeletionService.js
```diff
@@ -0,0 +1,80 @@
+const moment = require('moment')
```
jsrobertson, line 1: Not sure on the filename.  Feels more like `appointmentDetailsService.js` would be more suitable as there's nothing related to deletion in this file.

jsrobertson (approved):

PR #1699, opened 2021-06-07
NN-3433 capitalise name in aria-label and update tests

natclamp-moj:


backend/routes/appointments/viewAppointmentsRouter.js
```diff
@@ -100,7 +100,11 @@ module.exports = ({ prisonApi, whereaboutsApi, logError }) => async (req, res) =
           html: getCourtDescription(),
         },
         {
-          html: `<a href="/appointment-details/${appointment.id}" class="govuk-link">View details</a>`,
+          html: `<a href="/appointment-details/${appointment.id}" class="govuk-link" aria-label="View details of ${
+            appointment.firstName[0]
```
jsrobertson, line 104: We have a `formatName(firstName, lastName)` function you can use or in this instance you could probably use `posessive(offenderName)`. Can show you if needs be.

jsrobertson (approved):

PR #1706, opened 2021-06-10
Nn 3417 hide pre post appointments

natclamp-moj:


backend/tests/viewAppointmentsController.test.js
```diff
@@ -390,6 +390,293 @@ describe('View appointments', () => {
     })
   })
 
+  describe('when there are video link appointments with pre and post appointments within the same booking', () => {
+    beforeEach(() => {
+      whereaboutsApi.getAppointments.mockReturnValue([
```
sp-watson, line 395: This is quite a large test - it could do with splitting up if we can. I suggest one test for each type of VLB - assuming they are testing different things.

backend/routes/appointments/viewAppointmentsRouter.js
```diff
@@ -53,6 +111,7 @@ module.exports = ({ prisonApi, whereaboutsApi, logError }) => async (req, res) =
 
   const appointmentsEnhanced = appointments
     .filter(appointment => (type ? appointment.appointmentTypeCode === type : true))
+    .filter(appointment => !VLPreAppIds.includes(appointment.id) && !VLPostAppIds.includes(appointment.id))
```
sp-watson, line 114: I think this can be achieved by just looking at the VLB booking type, then we perhaps don't need to store the id's

backend/routes/appointments/viewAppointmentsRouter.js
```diff
@@ -33,16 +33,74 @@ module.exports = ({ prisonApi, whereaboutsApi, logError }) => async (req, res) =
...
+  const videoLinkAppointmentIds = videoLinkAppointmentsWithTimes.map(videoLinkAppointment => videoLinkAppointment.id)
 
   const videoLinkAppointmentResponse = await whereaboutsApi.getVideoLinkAppointments(
     res.locals,
     videoLinkAppointmentIds
   )
+
   const videoLinkAppointments = (videoLinkAppointmentResponse && videoLinkAppointmentResponse.appointments) || []
   const videoLinkAppointmentsMadeByTheCourt = videoLinkAppointments.filter(appointment => appointment.madeByTheCourt)
+  const videoLinkAppointmentsWithTimesAndBookingIds = videoLinkAppointmentsWithTimes.map(appointment => {
```
sp-watson, line 46: Can we move all this to a function (or 2) - it makes the function hard to follow.

I have a suggestion - can we not do all this as part of a mapping function when making `appointmentsEnhanced` below. It would be done after the filtering and would look for VLBs. If one was found then it would call a function to calculate the `appointmentTimes`. Not sure how easy it would be though - we can go through it together to see if it is feasible or better.

backend/routes/appointments/viewAppointmentsRouter.js
```diff
@@ -33,16 +33,74 @@ module.exports = ({ prisonApi, whereaboutsApi, logError }) => async (req, res) =
...
+    videoLinkAppointments.forEach(videoLinkApp => {
+      if (videoLinkApp.appointmentId === appointment.id) {
+        appointmentCopy.bookingId = videoLinkApp.bookingId
+      }
+    })
+    return appointmentCopy
+  })
+
+  const videoLinkAppointmentBookingIds = videoLinkAppointments.map(VLAppointment => VLAppointment.bookingId)
+  const VLAppointmentsWithPrePostCount = videoLinkAppointmentBookingIds.reduce((acc, curr) => {
```
sp-watson, line 57: We use lowercase first letters for our consts.

sp-watson (changes requested):

PR #1707, opened 2021-06-11
NN-3449: Remove whereabouts create appointment feature flag

sp-watson:
Now we can delete recurring appointments, It will always use whereabouts


JamieDraperUK (approved):
LGTM

PR #1708, opened 2021-06-11
Nn 3417 hide vlb prepost appointments

natclamp-moj:


backend/routes/appointments/viewAppointmentsRouter.js
```diff
@@ -51,10 +51,52 @@ module.exports = ({ prisonApi, whereaboutsApi, logError }) => async (req, res) =
...
+      otherAppointment =>
+        otherAppointment.bookingId === appointment.bookingId && otherAppointment.hearingType === 'POST'
+    )
+    if (preAppointment) {
+      appCopy.displayStartTime = preAppointment.startTime
+    }
+    if (postAppointment) {
+      appCopy.displayEndTime = postAppointment.endTime
+    }
+    return appCopy
```
sp-watson, line 82: I think it is best to avoid adding new variables if possible. Could you not return a structure with the start/endTimes reassigned, e.g.
"
return {
  ...appointment,
  startTime: displayStartTime || startTime,
  endTime: displayEndTime || endTime,
}
"

..you would then not need to do the "displayEndTime || endTime" further down, which adds some confusion.

backend/routes/appointments/viewAppointmentsRouter.js
```diff
@@ -51,10 +51,52 @@ module.exports = ({ prisonApi, whereaboutsApi, logError }) => async (req, res) =
...
+  }
+
   const appointmentsEnhanced = appointments
     .filter(appointment => (type ? appointment.appointmentTypeCode === type : true))
+    .map(appointment => addBookingIdAndHearingTypesToVLBAppointments(appointment))
```
sp-watson, line 87: This is good work - there is a nice separation of concerns. Probably better than what i would have done (which its put it all in 1 function)

backend/routes/appointments/viewAppointmentsRouter.js
```diff
@@ -51,10 +51,52 @@ module.exports = ({ prisonApi, whereaboutsApi, logError }) => async (req, res) =
       court: appointment.court,
     }))
 
+  const addBookingIdAndHearingTypesToVLBAppointments = appointment => {
+    const appCopy = { ...appointment }
+    if (appointment.appointmentTypeCode === 'VLB') {
+      videoLinkAppointments.forEach(vlAppointment => {
+        if (vlAppointment.appointmentId === appointment.id) {
+          appCopy.hearingType = vlAppointment.hearingType
+          appCopy.bookingId = vlAppointment.bookingId
```
sp-watson, line 60: Probably best to avoid reference to a copy - you could just return the existing structure with extra data (as i show in the comment below) right here. You would also change the return statement at the bottom to use the "appointment" variable.

sp-watson (commented):
Good work! Just a couple of comments, both of them about avoiding the copy objects

sp-watson (approved):
Looks good

PR #1709, opened 2021-06-15
DCS-1007 Prevent users from specifying duration of pre/post appointments 

andrewrlee:
Removing the ability to specify pre/post appointment duration - these will now always be 15 mins long instead.


GurnankCheema (approved):
LGTM

PR #1718, opened 2021-06-21
NN-3462 Package updates - eslint related

jsrobertson:
Update eslint related packages and fix new warnings and errors.

Also fixes pre commit related tasks which weren't running.


yasinmustafa (approved):
LGTM

PR #1721, opened 2021-06-22
NN-3279 Show 'Life imprisonment' next to the release date if the offender is in for Life and has no release date

yasinmustafa:
https://user-images.githubusercontent.com/3595969/122889227-8487fa80-d33a-11eb-9a6f-de919597f4d8.mp4


backend/controllers/prisonerProfile/prisonerSentenceAndRelease.js
```diff
@@ -28,15 +28,20 @@ module.exports = ({ prisonerProfileService, prisonApi, systemOauthClient }) => a
   const sentenceAdjustments = sentenceAdjustmentsViewModel(sentenceAdjustmentsData)
   const courtCases = courtCasesViewModel({ courtCaseData, sentenceTermsData, offenceHistory })
 
+  const determineLifeSentence = async () => {
+    const prisonerDetails = await prisonApi.getPrisonerDetails(res.locals, offenderNo)
```
yasinmustafa, line 32: This call to `getPrisonerDetails` returns is a little bit heavy and returns a lot of info - so kept it separate and it only runs in a specific circumstance when no release date is defined (so should run rarely)

One for discussion with the dev team - we could potentially improve performance by using the ElasticSearch `prisonerSearchService` rather than the `prisonApi`

sp-watson: Yes, this is covered by a ticket which is currently in Initial Elab: https://dsdmoj.atlassian.net/browse/NN-3272

sp-watson (approved):

PR #1723, opened 2021-06-23
NN-3185: Use reg exps for time period in int tests that use the current date/time

sp-watson:


integration-tests/integration/whereabouts/selectResidentialLocation.spec.js
```diff
@@ -1,16 +1,5 @@
 const moment = require('moment')
 
-// copy of { getCurrentPeriod } from backend/utils/.js
```
JamieDraperUK, line 3: This is duplicated in /selectActivityLocation.spec.js  too

JamieDraperUK (dismissed):

integration-tests/integration/whereabouts/selectResidentialLocation.spec.js
sp-watson: Argh - I did the hard work and left the unused function in.. I'll commit that now

JamieDraperUK (approved):

PR #1725, opened 2021-06-23
NN-3279 use offender search to determine if the prisoner has a life sentence term

yasinmustafa:
The original implementation was relying on a prison-api call to determine if the offender had a Life term, however that didnt quite work in all scenarios. It needed some extra derivation on the sentence terms - after a slack chat with other dps devs it turns out the prisoner search service does this derivation https://mojdt.slack.com/archives/C012FNNQ4A3/p1623854264294400

So changed it to call the elastic  prisoner-search-service which returns a flag which fits our purpose.

### DIFF
--- a/backend/controllers/prisonerProfile/prisonerQuickLook.js
+++ b/backend/controllers/prisonerProfile/prisonerQuickLook.js
@@ -16,13 +16,12 @@ const trackEvent = (telemetry, username, activeCaseLoad) => {
 }
 
 const captureErrorAndContinue = apiCall =>
-    [{}]
-  // new Promise(resolve => {
-  //   apiCall.then(response => resolve({ response })).catch(error => {
-  //     log.error(error)
-  //     resolve({ error: true })
-  //   })
-  // })
+  new Promise(resolve => {
+    apiCall.then(response => resolve({ response })).catch(error => {
+      log.error(error)
+      resolve({ error: true })
+    })
+  })
 
 const extractResponse = (complexData, key) => {
   if (!complexData || complexData.error) return null
@@ -82,7 +81,6 @@ module.exports = ({ prisonerProfileService, prisonApi, telemetry, offenderSearch
     nextVisitResponse,
     visitBalancesResponse,
     todaysEventsResponse,
-    prisonerDetailsResponse,
   ] = await Promise.all(
     [
       prisonerProfileService.getPrisonerProfileData(res.locals, offenderNo, username),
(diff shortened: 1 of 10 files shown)


backend/controllers/prisonerProfile/prisonerQuickLook.js
```diff
@@ -132,6 +125,18 @@ module.exports = ({ prisonerProfileService, prisonApi, telemetry }) => async (re
 
   trackEvent(telemetry, username, activeCaseLoad)
 
+  const extractLifeImprisonmentStatus = async () => {
+    const [prisonerDetailsResponse] = await Promise.all(
```
yasinmustafa, line 129: I kept this call separate from the other api calls (calls the elstic prisoner search service).. as it is only used to determine if the offender is on a life sentence (at the moment). Therefore only runs in the specific scenario when it's needed.

Although, it does return a lot of the data already on the quicklook tab and is faster than calling the prison-api. So we could refactor to use this call to populate most (maybe all) of the data

backend/controllers/prisonerProfile/prisonerQuickLook.js
```diff
@@ -132,6 +125,18 @@ module.exports = ({ prisonerProfileService, prisonApi, telemetry }) => async (re
 
   trackEvent(telemetry, username, activeCaseLoad)
 
+  const extractLifeImprisonmentStatus = async () => {
```
sp-watson, line 128: You don't need to use Promise.all here

backend/controllers/prisonerProfile/prisonerQuickLook.js
```diff
@@ -158,7 +163,7 @@ module.exports = ({ prisonerProfileService, prisonApi, telemetry }) => async (re
               sentenceData.sentenceDetail &&
               sentenceData.sentenceDetail.releaseDate &&
               moment(sentenceData.sentenceDetail.releaseDate).format('D MMMM YYYY')) ||
-            extractLifeImprisonmentStatus(prisonerDataResponse, prisoner, unableToShowDetailMessage),
+            (await extractLifeImprisonmentStatus()),
```
sp-watson, line 166: This function name is a bit misleading - actually it gets the Imp Status label

sp-watson (approved):

backend/controllers/prisonerProfile/prisonerQuickLook.js
yasinmustafa: Thanks Steve - good spot, done

backend/controllers/prisonerProfile/prisonerQuickLook.js
yasinmustafa: Agreed - renamed

sp-watson (approved):

PR #1726, opened 2021-06-23
NN-3478: implements template for offender neurodiversity summary information

ake3mio:
This PR:
- adds the UI and boilerplate for the neurodiversity summary accordion on the prisoner api
- Updates the nvmrc. node version to match the package json node major version

**Notes**

This work currently sits behind a temporary feature flag which can be toggled by setting an env called `ESWE_ENABLED`. Valid values for `ESWE_ENABLED` are `true` or `false`.

The feature flag is enabled for integration tests, but will not display in dev until the thirdparty api is ready so that we can integrate with it. For now mock data shown when the feature flag is enabled.

**Designs**
Design_for_Neurodiversity_accordion_on_DPS_profile_v4

**Feature flag not set or disabled**
Screenshot 2021-06-23 at 11 57 28

**Feature flag set to true and no data**
Screenshot 2021-06-23 at 09 04 57

**Feature flag set to true with data**
Screenshot 2021-06-23 at 13 11 35


backend/api/curiousApi.js
```diff
@@ -0,0 +1,47 @@
...
+  },
+]
+
+const curiousApiFactory = _ => ({
+  getLearnerProfiles() {
+    return Promise.resolve(dummyLearnerProfiles)
+  },
+})
+
+module.exports = { curiousApiFactory, dummyLearnerProfiles }
```
sp-watson, line 47: I don't think you need to expose dummyLearnerProfiles. It is going to cause problems down the line, and we don't generally have unit tests for Apis - we expect the integration tests to do that.

backend/api/curiousApi.test.js
```diff
@@ -0,0 +1,12 @@
+const { curiousApiFactory, dummyLearnerProfiles } = require('./curiousApi')
```
sp-watson, line 1: Not sure this is need - see comment above

backend/services/esweService.js
```diff
@@ -0,0 +1,45 @@
+const { app } = require('../config')
+
+/**
+ * Education skills and work experience (ESWE)
+ */
+class EsweService {
```
sp-watson, line 6: Any reason why this starts with a capital letter? We generally start with lowercase letters in this code base

backend/services/esweService.js
```diff
@@ -0,0 +1,45 @@
...
+  #curiousApi = null
```
sp-watson, line 7: Interesting - is this object style? The problem is.. we don't use these constructs anywhere else in our code base .. it would be interesting to know what this might give us

sp-watson (commented):
Looks good, but there are a couple of inconsistencies with the existing code base that need a discussion.

backend/services/esweService.js
sp-watson: We have agreed that TypeScript is the way forward, and this is a step in that direction.

sp-watson (approved):

PR #1727, opened 2021-06-24
NN-3462 Update prettier and apply associated code style changes

jsrobertson:
Updated prettier from 1 to 2 which has introduced a couple of changes.  Main one being is that parenthesis are now required for single params as mentioned [here](https://prettier.io/docs/en/options.html#arrow-function-parentheses) 

> At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the default value for the option.


yasinmustafa (approved):
LGTM!

PR #1729, opened 2021-06-25
NN-3389: Adds typescript support and moves curious related code to ts

ake3mio:
This PR introduces typescript to the project


jsrobertson (approved):
Lets do it. Nice one.

PR #1737, opened 2021-06-30
NN-3479: integrates with initial curious endpoints and migrates more code to typescript

ake3mio:
This PR:

- Migrates more code to typescript 
- Implements an initial integration with the curious api
- Enables eswe in the dev environment


backend/api/curious/curiousApi.test.ts
```diff
@@ -1,18 +1,198 @@
-import CuriousApi, { dummyLearnerProfiles, dummyLearnerLatestAssessments } from './curiousApi'
+import nock from 'nock'
```
sp-watson, line 1: Just noticed you have put the tests in the same dir as the source. We generally don't do that, so i would prefer to keep it like that unless there is a good reason (you probably have one).

sp-watson (approved):

PR #1760, opened 2021-07-07
NN-3527 Use prisonstaffhubclient token when making requests to the prisoner search service to get prisoner details

yasinmustafa:
In addition to this code change the GLOBAL_SEARCH role will need to be added to prisonstaffhubclient in auth


backend/tests/prisonerQuickLook.test.js
```diff
@@ -67,7 +68,16 @@ describe('prisoner profile quick look', () => {
 
     offenderSearchApi.getPrisonersDetails = jest.fn().mockResolvedValue([])
 
-    controller = prisonerQuickLook({ prisonerProfileService, prisonApi, telemetry, offenderSearchApi, logError })
+    systemOauthClient.getClientCredentialsTokens = jest.fn().mockReturnValue({})
```
JamieDraperUK, line 71: Need to test that the systemOauthClient gets called

JamieDraperUK, line 71: Need to test that the systemOauthClient gets called

JamieDraperUK (dismissed):
LGTM - missing unit tests

backend/tests/prisonerQuickLook.test.js
yasinmustafa: Thanks Jamie, agreed added tests for this now

JamieDraperUK (approved):

PR #1762, opened 2021-07-13
Nn 3477 functional skills updates

natclamp-moj:
This PR:
- fixes a bug that meant if the offender was in curious but had no assessments added, for the English/Welsh assessment, the label would be "English/Welsh" and not "English" as it was before.
- Wraps the getLearnerLatestAssessments function in a try block, so that if there is an error with the API call, it's caught and the error code checked - if it's a certain type, with errorCode VC500, this means the prisoner couldn't be found in Curious. This is logged and then an empty assessments summary is shown. Otherwise, the error is logged and the error content is shown instead of the summary
- The error content is updated to match the current content used by DPS Core

Example of what's shown if a prisoner is not in Curious:
scrnli_13_07_2021_13-59-39

Example of what's shown if there's an error calling the API/ error returned from the API that's not due to a missing prisoner:
scrnli_13_07_2021_14-07-23

Example of a prisoner with all assessment data:
scrnli_13_07_2021_14-10-24


backend/services/esweService.ts
```diff
@@ -105,72 +105,102 @@ export default class EsweService {
...
+          ],
+          digiLit: [
+            {
+              label: 'Digital Literacy',
+              value: 'Awaiting assessment',
+            },
+          ],
+        }
+      } else {
+        log.warn(`Failed to get latest learning assessments. Reason: ${e}`)
```
sp-watson, line 201: Does this not need to be a log.error? I suspect we need to know when users are not seeing the data.

backend/services/esweService.ts
```diff
@@ -105,72 +105,102 @@ export default class EsweService {
...
-        },
-      }
-      if (Array.isArray(functionalSkillLevels) && functionalSkillLevels.length > 0) {
-        const { qualifications } = functionalSkillLevels[0]
-        const learnerAssessment = qualifications
-          .filter((functionalSkillLevel) => functionalSkillLevel.qualification.qualificationType === qualificationType)
-          .sort(compareByDateDesc)[0]
-
-        return learnerAssessment || emptyAssessment
+    let skillLevels: Record<string, unknown> = null
```
sp-watson, line 108: Personally, I would avoid having this variable. Probably the easiest solution is add a return statement at the end of the try block. I know it will duplicate the createFlaggedContent() call, but that is done elsewhere in this file.

backend/services/esweService.ts
```diff
@@ -105,72 +105,102 @@ export default class EsweService {
...
-      english: createSkillAssessmentSummary(englishGrade),
-      maths: createSkillAssessmentSummary(mathsGrade),
-      digiLit: createSkillAssessmentSummary(digitalLiteracyGrade),
+      skillLevels = {
+        english: createSkillAssessmentSummary(englishGrade),
+        maths: createSkillAssessmentSummary(mathsGrade),
+        digiLit: createSkillAssessmentSummary(digitalLiteracyGrade),
+      }
+    } catch (e) {
+      if (e.response?.body?.errorCode === 'VC500') {
```
sp-watson, line 178: I notice there is another Curious API call in this file `getLearnerProfiles` - should errors from that be checked as well?

backend/services/esweService.ts
```diff
@@ -105,72 +105,102 @@ export default class EsweService {
...
+        log.info(`Offender record not found in Curious.`)
+        skillLevels = {
```
sp-watson, line 180: Not that important, but it might be worth moving this to a const at the top of this file with a name like `DEFAULT_SKILL_LEVELS`?

sp-watson (approved):

backend/services/esweService.ts
ake3mio: @sp-watson something similar will be picked up on: https://dsdmoj.atlassian.net/browse/NN-3499

backend/services/esweService.ts
```diff
@@ -105,72 +126,82 @@ export default class EsweService {
...
+    } catch (e) {
+      if (e.response?.body?.errorCode === 'VC500') {
+        log.info(`Offender record not found in Curious.`)
+        return createFlaggedContent(DEFAULT_SKILL_LEVELS)
+      }
+      log.error(`Failed to get latest learning assessments. Reason: ${e.message}`)
     }
-
-    return createFlaggedContent(functionalSkillLevels)
+    return createFlaggedContent(skillLevels)
```
ake3mio, line 205: Looking good @natclamp-moj . I think this function has gained the 'huge' status though.

Perhaps it could be broken down a bit to improve readability and help with maintenance. What do you think?

backend/tests/esweService.test.ts
```diff
@@ -293,6 +293,32 @@ describe('Education skills and work experience', () => {
...
+        const error = {
+          response: {
+            body: {
+              errorCode: 'VC500',
+            },
+          },
+        }
+        jest.spyOn(app, 'esweEnabled', 'get').mockReturnValue(true)
+        getLearnerLatestAssessmentsMock.mockRejectedValue(error)
+        const expectedResult = {
```
ake3mio, line 307: It might be best to to export `DEFAULT_SKILL_LEVELS` and just use that.

views/prisonerProfile/prisonerWorkAndSkills/prisonerWorkAndSkills.njk
```diff
@@ -6,7 +6,7 @@
 
 {% set title = "Work and Skills" %}
 {% set containerClasses = "prisoner-profile" %}
-{% set missingLevel = "Missing data" %}
+{% set apiErrorMessage = "Unable to show this detail." %}
```
ake3mio, line 9: **Non blocking**
I wonder if a view need to know about an api?
The data wend from api -> service -> routeHandler -> browser.

Perhaps the view should stay decoupled from the concept of `api`. 

Maybe something like `fallbackMessage` or `errorMessage`?

backend/services/esweService.ts
ake3mio: It might be could to document these codes and what they mean @natclamp-moj 
Are there other curious codes?

backend/services/esweService.ts
ake3mio: Yes, it might be more clear to return it.
The function can be broken down to decrease the overall size and complexity also

ake3mio (approved):
👍

PR #1763, opened 2021-07-14
NN-3545 Do not display the feedback banner when a page is printed

yasinmustafa:


jsrobertson (changes requested):
Better to use `govuk-!-display-none-print`

Also needs to be applied to the .njk version of the footer too.

jsrobertson (approved):

PR #1764, opened 2021-07-15
Nn 3548 refactor functional skills

natclamp-moj:
This PR:
- refactors the esweService getLatestLearnerAssessments function, removing unnecessary sorting on the API results and removes replication
- renames error variables
- adapts to the new error response from megaNexus regarding a record not found
- includes a new unit test, and corrects dummy data used
- adapts integration test for when the prisoner is not in Curious

### DIFF
--- a/backend/services/esweService.ts
+++ b/backend/services/esweService.ts
@@ -168,7 +168,7 @@ export class EsweService {
         functionalSkillLevels: curious.LearnerLatestAssessment[],
         qualificationType: AssessmentQualificationType
       ): curious.LearnerAssessment => {
-        const emptyAssessment = {
+        const emptyAssessment: curious.LearnerAssessment = {
           qualification: {
             qualificationType,
           },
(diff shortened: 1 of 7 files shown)


backend/services/esweService.ts
```diff
@@ -55,10 +57,42 @@ const compareByDate = (dateA: Moment, dateB: Moment, descending = true) => {
...
+      value: establishmentName,
+    },
+  ]
+}
+
 /**
  * Education skills and work experience (ESWE)
  */
-export default class EsweService {
+export class EsweService {
```
ake3mio, line 95: **Not blocking**
Was there a reason this was changed from a default export?

ake3mio (approved):
LGTM

backend/services/esweService.ts
natclamp-moj: I was under the impression if you had a default export you couldn't export other things separately - not sure why, I'll swap it back

ake3mio (approved):
👍

PR #1774, opened 2021-07-28
Pat 1950 soc api

lbennett-moj:
Switching to new API for SOC member lookup, this was previously accessing an endpoint on the ui service


backend/services/prisonerProfileService.ts
```diff
@@ -172,9 +172,9 @@ export default ({
       showPathfinderReferButton: Boolean(!pathfinderDetails && isPathfinderReadWriteUser),
       pathfinderReferUrl: pathfinderUrl && `${pathfinderUrl}refer/offender/${offenderNo}`,
       canViewSocLink: socEnabled && canViewSocLink,
-      socProfileUrl: socEnabled && socUrl && socDetails && `${socUrl}/nominal/${String((socDetails as any).id)}`,
```
petergphillips, line 175: this looks a bit naff - we normally try to exclude `/` from the config url so that the urls look nicer

backend/api/socApi.ts
```diff
@@ -2,7 +2,7 @@ export const socApiFactory = (client) => {
   const processResponse = () => (response) => response.body
   const get = (context, url) => client.get(context, url).then(processResponse())
   const getSocDetails = (context, offenderNo, socEnabled) =>
-    socEnabled ? get(context, `/api/offender/${offenderNo}`) : Promise.resolve(null)
+    socEnabled ? get(context, `/soc/nominal/nomsId/${offenderNo}`) : Promise.resolve(null)
```
petergphillips, line 5: odd choice here - new apis should really be using kebab-case instead

backend/config.ts
```diff
@@ -110,7 +110,8 @@ export const apis = {
     timeoutSeconds: toNumber(process.env.API_ENDPOINT_TIMEOUT_SECONDS) || 30,
   },
   soc: {
-    url: process.env.SOC_URL || '',
+    url: process.env.SOC_API_URL || '',
```
petergphillips, line 113: would have been clearer to define as `api_url` here instead

helm_deploy/values-preprod.yaml
```diff
@@ -42,7 +42,8 @@ env:
   PATHFINDER_ENDPOINT_API_URL: https://preprod-api.pathfinder.service.justice.gov.uk/
   PIN_PHONES_URL: https://pin-phone-monitor-qa.prison.service.justice.gov.uk/
   REDIS_ENABLED: true
-  SOC_URL: https://manage-soc-cases-preprod.hmpps.service.justice.gov.uk
+  SOC_UI_URL: https://manage-soc-cases-preprod.hmpps.service.justice.gov.uk/
```
petergphillips, line 45: iirc I think we're trying to have new urls without the trailing slash so that they read nicer - am I correct @JamieDraperUK ?

petergphillips (approved):
LGTM

backend/services/prisonerProfileService.ts
lbennett-moj: i went with the majority to keep it consistent

lbennett-moj: you are right - i meant to take that out

backend/api/socApi.ts
lbennett-moj: noted - will discuss

helm_deploy/values-preprod.yaml
lbennett-moj: I went with the majority here

PR #1777, opened 2021-08-03
NN-3581 fix regression in the health check code that provides the bui…

JamieDraperUK:
…ld info

### DIFF
--- a/backend/services/healthCheck.ts
+++ b/backend/services/healthCheck.ts
@@ -24,7 +24,6 @@ const getBuild = () => {
 
 const addAppInfo = (result) => {
   const buildInformation = getBuild()
-  console.error({ buildInformation })
   const buildInfo = {
     uptime: process.uptime(),
     build: buildInformation,
--- a/backend/services/healthCheck.ts
+++ b/backend/services/healthCheck.ts
@@ -15,9 +15,8 @@ const gatherCheckInfo = (total, currentValue) => ({ ...total, [currentValue.name
 
 const getBuild = () => {
   try {
-    return JSON.parse(fs.readFileSync(path.join(__dirname, '../../../build-info.json'), 'utf-8'))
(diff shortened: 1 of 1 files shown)


backend/services/healthCheck.ts
```diff
@@ -13,9 +15,9 @@ const gatherCheckInfo = (total, currentValue) => ({ ...total, [currentValue.name
 
 const getBuild = () => {
   try {
-    // eslint-disable-next-line import/no-unresolved,global-require
-    return require('../../build-info.json')
+    return JSON.parse(fs.readFileSync(path.join(__dirname, '../../../build-info.json'), 'utf-8'))
   } catch (ex) {
+    console.error(ex)
```
ake3mio, line 20: Should we be using the logger here?

backend/services/healthCheck.ts
```diff
@@ -13,9 +15,9 @@ const gatherCheckInfo = (total, currentValue) => ({ ...total, [currentValue.name
 
 const getBuild = () => {
   try {
-    // eslint-disable-next-line import/no-unresolved,global-require
-    return require('../../build-info.json')
+    return JSON.parse(fs.readFileSync(path.join(__dirname, '../../../build-info.json'), 'utf-8'))
```
ake3mio, line 18: *non blocking*
Perhaps using `process.cwd()` might be better over `__dirname`. It would make sure that the file is always located from wherever the `npm run start` command was run on the path.

ake3mio (approved):

ake3mio (approved):

backend/services/healthCheck.ts
JamieDraperUK: Good shout!

backend/services/healthCheck.ts
JamieDraperUK: Not really needed tbh

ake3mio (approved):

PR #1801, opened 2021-09-03
Eswe 44 courses and qualifications detail

natclamp-moj:
This PR:
- alters curious API call to add ability to use queries
- create new function in esweService to get full courses and qualifications information 
- new controller and view for details page

scrnli_03_09_2021_15-46-05


backend/controllers/prisonerProfile/prisonerCoursesQualificationsDetails.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    if (!app.esweEnabled) {
+      return res.redirect(`/prisoner/${offenderNo}`)
+    }
+
+    try {
+      const [prisonerDetails, coursesAndQualifications]: [PrisonerDetails, coursesAndQuals] = await Promise.all([
+        prisonApi.getDetails(res.locals, offenderNo),
+        esweService.getLearnerEducationFullDetails(offenderNo),
+      ])
```
sp-watson, line 35: I think we need `logErrorAndContinue` here instead of a `try {}` block as it prevents a failure in one of these calls from stopping the whole prisoner profile page from being shown.

backend/api/curious/types/index.d.ts
```diff
@@ -414,6 +414,56 @@ declare namespace curious {
     historicalCoursesPresent?: boolean
   }
 
+  /**
+   *
+   * @export
+   * @interface LearnerEducationFullDetails
+   */
+  interface LearnerEducationFullDetails {
```
sp-watson, line 422: [Not to fix now - just a design suggestion]

IMO this isn't quite the right place for this definition as it isn't used by the Curious API. Given that it is created by EsweService i would define and export it there.

However i haven't looked at other type definitions here - there may be more examples of this, in which case it probably isn't worth fixing any time soon.

backend/controllers/prisonerProfile/prisonerCoursesQualificationsDetails.ts
```diff
@@ -0,0 +1,49 @@
...
+      agencyId: string
+      assignedLivingUnitId: number
+    }
+
+    type coursesAndQuals = {
+      enabled: boolean
+      content: curious.LearnerEducationFullDetails[]
+    }
+
+    if (!app.esweEnabled) {
```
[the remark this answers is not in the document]
natclamp-moj: I left this out in a previous education ticket and John S found that if he went directly to the URL he could bypass the flag completely (you wouldn't see any data because the EsweService stuff is also behind the flag) but you could get to an essentially blank page. So it's just in there to stop that. Will be removed when we take the flag out.

backend/api/curious/curiousApi.ts
```diff
@@ -18,9 +18,38 @@ export default class CuriousApi {
       .then((response) => response.body)
   }
 
-  getLearnerEducation(context: ClientContext, nomisId: string): Promise<curious.LearnerEducation> {
+  getLearnerEducation(
+    context: ClientContext,
+    nomisId: string,
+    sort?: string,
```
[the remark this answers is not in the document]
natclamp-moj: Not sure all of them will be used honestly, I'll take out the ones I'm not sure about

backend/controllers/prisonerProfile/prisonerCoursesQualificationsDetails.ts
natclamp-moj: So this is actually a completely separate page, not part of the prisoner profile - I also did have the logErrorAndContinue there in a map but it causes an almighty typescript error

backend/api/curious/types/index.d.ts
natclamp-moj: I've put most of my definitions in there as I was unsure where else to put them, so it's a bigger task - though I can put something in the backlog to sort this out when the MVP is done

backend/api/curious/curiousApi.test.ts
```diff
@@ -52,6 +52,176 @@ describe('curiousApi', () => {
       const actual = await curiousApi.getLearnerEducation({ access_token: accessToken }, nomisId)
       expect(actual).toEqual(dummyEducations)
     })
+    it('should return the expected response data with filter in query', async () => {
+      const nomisId = dummyEducations.content[1].prn
+      const { establishmentId } = dummyEducations.content[1]
+      nock(hostname)
```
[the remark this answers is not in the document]
natclamp-moj: I swapped it out as there was something I was trying to solve, but it looks like I solved the issue and then didn't change it back... changing

backend/controllers/prisonerProfile/prisonerCoursesQualificationsDetails.ts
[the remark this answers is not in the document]
sp-watson: Ok, fair enough

backend/controllers/prisonerProfile/prisonerCoursesQualificationsDetails.ts
sp-watson: ah yes - i remember now you come to mention it! it fine as is then

backend/api/curious/types/index.d.ts
sp-watson: ok - it isn't a disaster, just a bit misleading. Not sure it is even worth a backlog ticket, purely because we would never get round to it!

sp-watson (approved):

PR #1802, opened 2021-09-03
ESWE-79 apply changes to fit with new api format and refactor ldd inf…

natclamp-moj:
This PR:
- matches changes to the new api spellings for LDD info
- refactors the LDD function so that html is only dealt with in the template files

### DIFF
--- a/views/prisonerProfile/macros/prisonerSummarySection.njk
+++ b/views/prisonerProfile/macros/prisonerSummarySection.njk
@@ -17,7 +17,13 @@
 
   {% for detail in array %}
 
-    {% if detail.html %}
+    {% if detail.ldd %}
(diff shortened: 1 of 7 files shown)


views/prisonerProfile/macros/prisonerSummarySection.njk
```diff
@@ -17,7 +17,13 @@
 
   {% for detail in array %}
 
-    {% if detail.html %}
+    {% if detail.ldd %}
```
sp-watson, line 20: Yes - like this, but...

We ought to give it a generic name instead of `ldd` (and `lddList`) - maybe just `detail.values`?

Can be fixed later though.

sp-watson (approved):

PR #1815, opened 2021-09-15
Eswe 48 work inside prison details

natclamp-moj:
This PR: creates the new work inside prison page and associated tests


backend/services/esweService.ts
```diff
@@ -391,4 +392,42 @@ export default class EsweService {
...
   }
+
+  async getWorkHistoryDetails(nomisId: string): Promise<workHistoryFullDetails> {
+    if (!app.esweEnabled) {
+      return createFlaggedContent(null)
+    }
+
+    try {
+      const context = await this.systemOauthClient.getClientCredentialsTokens()
+      const oneYearAgo = moment().subtract(1, 'year').format('YYYY-MM-DD')
```
sp-watson, line 403: Not a blocker, but the code to get the work history to show shouldn't be duplicated code.

This code and the code above have the same reason for change - if this changes the other one must. Therefore they should be in one place.

backend/utils/nunjucksSetup.ts
```diff
@@ -147,6 +154,10 @@ export default (app) => {
     return value || specifiedText || '--'
   })
 
+  njkEnv.addFilter('sortableDate', (date) => {
```
sp-watson, line 157: Is this used anywhere?

backend/utils/nunjucksSetup.ts
```diff
@@ -47,6 +47,13 @@ export default (app) => {
 
   njkEnv.addFilter('formatDate', (value, format) => (value ? moment(value).format(format) : null))
 
+  njkEnv.addFilter('checkValidAndFormatDate', (value, format) => {
```
sp-watson, line 50: Not sure about the name - maybe `getSortableDate` would make more sense?

Actually looking at where it is used, it is only called if there isn't a date - so maybe `defaultSortDate`?

views/prisonerProfile/prisonerWorkAndSkills/prisonerWorkInsidePrisonDetails.njk
```diff
@@ -0,0 +1,96 @@
...
+    classes: "govuk-!-display-none-print"
+  }) }}
+{% endblock %}
+
+{% set rows = [] %}
+{% for job in workDetailsList %}
+  {% set rows = (rows.push([
+    { text: job.role },
+    { text: job.location },
+    { text: job.startDate | formatTimestampToDate, attributes: { "data-sort-value": job.startDate | formatDate("X") } },
```
sp-watson, line 30: What does X mean? Is this required is the startDate is always set?

sp-watson (approved):

views/prisonerProfile/prisonerWorkAndSkills/prisonerWorkInsidePrisonDetails.njk
natclamp-moj: It's the way I've seen it's done elsewhere in the repo - it's a unix format - thought I'd stay consistent with existing code. The date sorting doesn't seem to work otherwise

backend/utils/nunjucksSetup.ts
natclamp-moj: Will rename this, defaultSortDate works - it just wouldn't sort when there was text in some cells and dates in others

PR #1821, opened 2021-09-17
DT-2461: ✨ Add in omic open case note warnings to amend case note

petergphillips:
Screenshot from 2021-09-17 14-18-55

### DIFF
--- a/static/js/caseNotes.js
+++ b/static/js/caseNotes.js
@@ -9,7 +9,7 @@ $(function () {
     if (!selectedTypeOption) return
 
     subTypeSelect.prop('disabled', true)
-    omicOpenWarnings.prop('style', 'display: none')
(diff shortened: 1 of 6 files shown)


static/js/caseNotes.js
```diff
@@ -9,7 +9,7 @@ $(function () {
     if (!selectedTypeOption) return
 
     subTypeSelect.prop('disabled', true)
-    omicOpenWarnings.prop('style', 'display: none')
+    if (omicOpenWarnings.length) omicOpenWarnings.prop('style', 'display: none')
```
petergphillips, line 12: apparently have to do a `.length` on the jquery object to see if it exists

petergphillips: which it won't do on the case notes tab

JamieDraperUK (approved):
LGTM

PR #1822, opened 2021-09-17
ESWE-47 add extra filter to current work

natclamp-moj:
This PR:
- adds an extra filter to the getCurrentWork function within the ESWE service in order to show only jobs that are in the current caseload

Also: applies the stringWithAbbreviationsProcessor function to an uppercase establishment name


JamieDraperUK (approved):
LGTM

PR #1824, opened 2021-09-20
ESWE-88 add temporarily withdrawn content

natclamp-moj:


backend/services/esweService.ts
```diff
@@ -290,7 +290,9 @@ export default class EsweService {
           )
           .sort(compareByDateAsc)
           .map((course) => ({
-            label: course.courseName,
+            label: course.completionStatus.includes('temporarily')
```
sp-watson, line 293: This is a bit "fluffy" - is there >1 completion status with that word in? I would prefer exact references unless MN know that you/we are checking for a particular word in the status.

I know that the why you have done it might mean we don't have to change anything if a status is added, but equally it might break it?!

sp-watson: Turns out the status is very wordy (more of a description), so this solution is as good as any

PR #1840, opened 2021-09-27
NN-3643: changing call back urls fo the prison staff hub

richardpopple:


backend/config.ts
```diff
@@ -58,7 +58,7 @@ export const apis = {
     url: process.env.OAUTH_ENDPOINT_URL || 'http://localhost:9090/auth/',
     ui_url: process.env.OAUTH_ENDPOINT_UI_URL || process.env.OAUTH_ENDPOINT_URL || 'http://localhost:9090/auth/',
     timeoutSeconds: toNumber(process.env.API_ENDPOINT_TIMEOUT_SECONDS) || 10,
-    clientId: process.env.API_CLIENT_ID || 'elite2apiclient',
+    clientId: process.env.API_CLIENT_ID || 'prisonapiclient',
```
richardpopple, line 61: We now use the the prisonapiclient in local development as that that is backed by the correct callback urls in the auth app

backend/sessionManagementRoutes.ts
```diff
@@ -94,13 +94,17 @@ export const configureRoutes = ({ app, tokenRefresher, tokenVerifier, homeLink }
...
-      res.redirect(`/login?${query}`)
+      res.redirect(`/sign-in?${query}`)
     })
   }
 
-  app.get('/login', loginMiddleware, remoteLoginIndex)
+  app.get('/sign-in', signInMiddleware, remoteSignInIndex)
 
-  app.get('/login/callback', (req, res, next) => {
+  app.get('/login/callback', (req, res) => {
```
richardpopple, line 103: redirect for any book marks

integration-tests/integration/prisonerProfile/prisonerCellHistory.spec.js
```diff
@@ -86,21 +86,15 @@ context('Prisoner cell history', () => {
         .cellDetailsLink()
         .should('contain.text', 'View details for location 1-02')
         .should('have.attr', 'href')
-        .and(
-          'include',
-          `/location-history?fromDate=2020-05-01T12:48:33&toDate=${moment().format(
-            'YYYY-MM-DDTHH:mm:ss'
-          )}&locationId=1&agencyId=MDI`
-        )
+        .and('contains', `/location-history?fromDate=2020-05-01T12:48:33&toDate=${moment().format('YYYY-MM-DDTHH:mm')}`)
```
richardpopple, line 89: Don't match to the second as this causes intermittent failures

integration-tests/integration/signIn/signIn.spec.js
```diff
@@ -14,72 +14,66 @@ context('Login functionality', () => {
...
+  it('Root (/) redirects to the auth sign in page if not signed in', () => {
+    cy.task('stubSignInPage')
     cy.visit('/')
     cy.url().should('include', 'authorize')
     cy.get('h1').should('contain.text', 'Sign in')
   })
 
-  it('Login page redirects to the auth login page if not logged in', () => {
-    cy.task('stubLoginPage')
-    cy.visit('/login')
```
sp-watson, line 26: I think we should keep a `/login` test, as it is still there (albeit just a redirect)

richardpopple: yeah - your right - I'll have a look at that.

sp-watson (dismissed):

sp-watson (approved):

PR #1842, opened 2021-09-27
NN-1867 court appearances

JamieDraperUK:
- Display prisoners due to leave to court for a given date


backend/api/prisonApi.ts
```diff
@@ -400,6 +409,12 @@ export const prisonApiFactory = (client) => {
   const getOffenderWorkHistory = (context, offenderNo, earliestEndDate) =>
     get(context, `/api/offender-activities/${offenderNo}/work-history?earliestEndDate=${earliestEndDate}`)
 
+  const getTransfers = (context, parameters: GetTransferParameters) =>
+    get(
+      context,
+      `/api/movements/transfers?agencyId=${parameters.agencyId}&fromDateTime=${parameters.fromDateTime}&toDateTime=${parameters.toDateTime}&releaseEvents=${parameters.releaseEvents}&courtEvents=${parameters.courtEvents}&transferEvents=${parameters.transferEvents}`
```
petergphillips, line 415: this looks really naff - would be better to use querystring.stringify or similar to do this automatically - see https://github.com/ministryofjustice/manage-hmpps-auth-accounts/blob/main/backend/api/prisonApi.js#L30

backend/controllers/whereabouts/scheduledMoves.ts
```diff
@@ -24,6 +21,98 @@ export default ({ prisonApi }) => {
...
         .sort(sortTextAlphabetically),
+      courtEvents,
+    })
+
+  const index = async (req, res) => {
+    const { userDetails } = req.session
+    const { activeCaseLoadId } = userDetails
+    const date = req.query?.date || moment().format('DD/MM/YYYY')
+
+    const movementReasons = await prisonApi.getMovementReasons(res.locals)
```
petergphillips, line 32: any reason for not using `Promise.all` here - would allow all three `await` events to be processed in parallel rather than having to wait for each in turn

backend/controllers/whereabouts/scheduledMoves.ts
```diff
@@ -24,6 +21,98 @@ export default ({ prisonApi }) => {
...
+    const transfers = await prisonApi.getTransfers(res.locals, {
+      fromDateTime: isoDateTimeStartOfDay(date, 'DD/MM/YYYY'),
+      toDateTime: isoDateTimeEndOfDay(date, 'DD/MM/YYYY'),
+      agencyId: activeCaseLoadId,
+      courtEvents: true,
+      transferEvents: true,
+      releaseEvents: true,
+    })
+
+    const offenderNumbers = [
```
petergphillips, line 44: won't you end up with duplicates here?  Would be good to therefore remove by putting into a set - maybe something like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#relation_with_array_objects?

petergphillips (approved):
LGTM

backend/controllers/whereabouts/scheduledMoves.ts
```diff
@@ -24,6 +21,98 @@ export default ({ prisonApi }) => {
...
+      ...transfers.courtEvents.map((ce) => ce.offenderNo),
+      ...transfers.transferEvents.map((te) => te.offenderNo),
```
sp-watson, line 46: How come we need to consider transferEvents and releaseEvents when only courtEvents are actually used?

backend/controllers/whereabouts/scheduledMoves.ts
```diff
@@ -24,6 +21,98 @@ export default ({ prisonApi }) => {
...
+
+    if (offenderNumbers.length === 0) {
+      return renderTemplate(res, {
+        date,
+        agencyDetails,
+        movementReasons,
+        courtEvents: [],
+      })
+    }
+
```
sp-watson, line 58: Consider moving the following logic to a function - something like getCourtEvents() - then we don't need the check above and it might make things a bit clearer?

backend/controllers/whereabouts/scheduledMoves.ts
```diff
@@ -24,6 +21,98 @@ export default ({ prisonApi }) => {
...
+    const searchResult = await offenderSearchApi.getPrisonersDetails(res.locals, offenderNumbers)
```
sp-watson, line 59: The use of generic names like `searchResult` and `prisonerProperty` make this a bit confusing IMO. And they are singular - but I guess there will be many?

backend/controllers/whereabouts/scheduledMoves.ts
```diff
@@ -24,6 +21,98 @@ export default ({ prisonApi }) => {
...
+          )
+        )
+      )) || []
+
+    const prisonerCommonShared = searchResult.map((details) => {
+      const alertCodes = details.alerts
+        .map((alert) => alert.alertCode)
+        .filter((alert) => relevantAlertsForTransfer.includes(alert))
+
+      const relevantAlertLabels = alertFlagLabels.filter((flag) =>
```
sp-watson, line 80: Won't `relevantAlertsForTransfer` just include codes in `alertFlagLabels`? So you just need to have another map above rather than 2 statements?!

backend/controllers/whereabouts/scheduledMoves.ts
```diff
@@ -24,6 +21,98 @@ export default ({ prisonApi }) => {
...
+        flag.alertCodes.some((code) => alertCodes.includes(code))
+      )
+
+      const property = prisonerProperty
```
sp-watson, line 84: Is there a better name than `property` (and `prisonerProperty`?) - property is an overloaded term - maybe `personalProperty` and `prisonerPersonalProperty`?

backend/controllers/whereabouts/scheduledMoves.ts
JamieDraperUK: Release event are coming next.

backend/controllers/whereabouts/scheduledMoves.ts
JamieDraperUK: Agreed. Once the release events done I'll start extracting the logic.

backend/controllers/whereabouts/scheduledMoves.ts
JamieDraperUK: Agreed. I'll improve the naming in the next PR

backend/controllers/whereabouts/scheduledMoves.ts
JamieDraperUK: Agreed.

backend/controllers/whereabouts/scheduledMoves.ts
JamieDraperUK: There are three data sets. Prisoner alerts, applicable alerts for transfers and the view model data that contains things like image location and correct css class. I'll try improve the readability in the next PR as this is a work in progress.

PR #1844, opened 2021-09-28
DT-2621: 🐛 Remove default case note type option on browser back

petergphillips:


views/amendCaseNote.njk
```diff
@@ -58,7 +58,7 @@
         </div>
     {% endfor %}
 
-    <form action="{{ postAmendment }}" method="post">
+    <form action="{{ postAmendment }}" method="post" autocomplete="off">
```
petergphillips, line 61: The select option wasn't being set at the time of the `ready` event, but later on so couldn't use the javascript to load the children.  According to https://stackoverflow.com/questions/8861181/clear-all-fields-in-a-form-upon-going-back-with-browser-back-button ended up with this solution

JamieDraperUK (approved):

PR #1852, opened 2021-09-29
NN-1867 add transfers and releases

JamieDraperUK:
- Display transfers and releases for a given date
- Centralise common njk template code
- Neaten up controller

Tests to confirm the API's are called correctly will be in the next PR


backend/controllers/whereabouts/scheduledMoves.ts
```diff
@@ -22,14 +22,63 @@ export default ({ prisonApi, offenderSearchApi }) => {
...
+    return prisonerDetailsForOffenderNumbers.map((details) => {
+      const personalProperty = personalPropertyForPrisoners
+        .flatMap((p) => p)
+        .filter((p) => p.prisonerNumber === details.prisonerNumber)
+        .map((p) => ({
+          containerType: p.containerType,
+          boxNumber: p.userDescription,
+        }))
+
+      return {
```
sp-watson, line 66: Just a note to not forget that we have typescript available now - we have the option to use types/classes. It might make some of this stuff a bit easier to read/maintain.

sp-watson (approved):

PR #1862, opened 2021-10-01
NN-1867 fix filtering

JamieDraperUK:
- Fix filtering
- Add in retry override


backend/controllers/whereabouts/scheduledMoves.ts
```diff
@@ -23,26 +26,21 @@ type ScheduledMovementDetails = {
...
-        .map((values) => ({
-          value: values.code,
-          text: values.description,
-        }))
-        .sort(sortTextAlphabetically),
+      scheduledTypes,
       courtEvents,
       releaseEvents,
       transferEvents,
+      showCourtAppearances: !scheduledType || scheduledType === 'Court',
```
sp-watson, line 41: Would it be possible to apply the scheduledType filter on the initial search `prisonApi.getTransfers` - that has booleans for each of them?!

sp-watson (approved):

PR #1868, opened 2021-10-06
NN-3625: rename prisonstaffhub

richardpopple:


package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "prisonstaffhub",
```
richardpopple, line 2: This I'm pretty sure will change the output roleName in insights, I will test when its on dev

backend/tests/prisonerFullImage.test.ts
```diff
@@ -46,7 +46,7 @@ describe('prisoner profile full image', () => {
   })
 
   it('should render with the correct back url if there is a referer', async () => {
-    const refererUrl = `//prisonStaffHubUrl/prisoner/${offenderNo}/personal`
+    const refererUrl = `//digitalPrisonServices/prisoner/${offenderNo}/personal`
```
sp-watson, line 49: Just checking - is this right? Should it not be `digitalPrisonServicesUrl`

sp-watson (approved):

PR #1877, opened 2021-10-08
ESWE-111 alter work inside prison

steverendell:


backend/tests/prisonerWorkInsidePrisonDetails.test.ts
```diff
@@ -15,34 +15,44 @@ describe('Prisoner work inside prison details controller', () => {
     getDetails: jest.fn(),
   }
   const esweService = {
```
sp-watson, line 17: Could maybe do with a test for pagination, although an int test is probably more useful for that

integration-tests/integration/prisonerProfile/prisonerWorkInsidePrisonDetails.spec.js
```diff
@@ -60,8 +86,10 @@ context('Work inside prison details page', () => {
 
     it('should show the correct content and number of results', () => {
```
sp-watson, line 87: I think we need an int test for pagination

PR #1900, opened 2021-10-19
NN-3717 add in request rate limiter

JamieDraperUK:
- Limit requests for a single page and IP address within a given threshold to avoid people spamming F5, creating more work for our backends. 

Longer term it would be good to have this at the Nginx level so it works for all our services.


backend/middleware/requestLimiter.ts
```diff
@@ -0,0 +1,33 @@
...
+import redis from 'redis'
+
+import config from '../config'
+
+// If more than 10 requests are made to the same url and ip address within 30 seconds
+// the rate limiter will kick in and the user will receive "Too many requests, please try again later."
+// and an HTTP 429 status code
+export default (maxConnections?) =>
+  new RateLimit({
+    ...initRedisStoreIfEnabled({
```
richardpopple, line 12: What happens if redis isn't enabled? Does it just keep a tally of requests for the server its on?

natclamp-moj (approved):

PR #1923, opened 2021-11-15
Nn 3740 feedback element adjudications history

natclamp-moj:


views/prisonerProfile/adjudicationHistory.njk
```diff
@@ -119,4 +119,10 @@
         {{ mojPagination(pagination) }}
     {% endif %}
 
+		<div>
```
sp-watson, line 122: BNTW we try to avoid using tabs on any of our files - the trouble is our code viewers are inconsistent when it comes to rendering them

sp-watson (approved):

PR #1928, opened 2021-11-19
Dt 2880 probation documents protected by caseload

andymarke:
To view probation document you now need to following
1) The correct role - this is unchanged
2) A caseload that matches the prisoner location - this is new

Therefore a POM can no longer view prisoners that have been released or in a another prisoner they don't have a caseload for.

This is a requested change from the MPC team alongside App Support


petergphillips (approved):
LGTM

PR #1933, opened 2021-11-24
CRS-534 adding calculate release dates button to prisoner profile page.

ldlharper:
Adding a calculate release dates button to the prisoner profile page.

We're in the middle of creating our pre prod environment now. There is a sub task to check the url in the dps helm config when we have the service running, we will do the same thing with production

The button won't appear on preprod or prod because no one has the calc role currently. We will use the role to act as a feature toggle.

### DIFF
--- a/backend/tests/prisonerProfileService.test.ts
+++ b/backend/tests/prisonerProfileService.test.ts
@@ -255,7 +255,7 @@ describe('prisoner profile service', () => {
         offenderNo: 'ABC123',
         offenderRecordRetained: undefined,
         showAddKeyworkerSession: false,
-        showCalculateRelseaseDates: false,
+        showCalculateReleaseDates: false,
         showReportUseOfForce: false,
         useOfForceUrl: '//useOfForceUrl/report/123/report-use-of-force',
         userCanEdit: false,
@@ -822,7 +822,7 @@ describe('prisoner profile service', () => {
 
           expect(getPrisonerProfileData).toEqual(
             expect.objectContaining({
-              showCalculateRelseaseDates: true,
+              showCalculateReleaseDates: true,
               calculateReleaseDatesUrl: 'http://crd-ui/?prisonId=ABC123',
             })
           )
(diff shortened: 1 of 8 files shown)


backend/tests/prisonerProfileService.test.ts
```diff
@@ -250,6 +255,7 @@ describe('prisoner profile service', () => {
         offenderNo: 'ABC123',
         offenderRecordRetained: undefined,
         showAddKeyworkerSession: false,
+        showCalculateRelseaseDates: false,
```
petergphillips, line 258: tpyo

helm_deploy/values-dev.yaml
```diff
@@ -62,4 +62,5 @@ env:
   PRISONS_WITH_MANAGE_ADJUDICATIONS_ENABLED: ""
   MANAGE_ADJUDICAIONS_URL: https://manage-adjudications-dev.hmpps.service.justice.gov.uk/
   PVB_URL: https://manage-a-prison-visit-dev.hmpps.service.justice.gov.uk/
+  CALCULATE_RELEASE_DATES_URL: https://calculate-release-dates-dev.hmpps.service.justice.gov.uk/
```
petergphillips, line 65: new urls shouldn't have a trailing slash otherwise paths look naff when calling them

helm_deploy/values-preprod.yaml
```diff
@@ -62,6 +62,7 @@ env:
   CURIOUS_URL: https://testservices.sequation.net/sequation-virtual-campus2-api/
   DISABLE_REQUEST_LIMITER: true
   PVB_URL: https://manage-a-prison-visit-preprod.hmpps.service.justice.gov.uk/
+  CALCULATE_RELEASE_DATES_URL: https://calculate-release-dates-preprod.hmpps.service.justice.gov.uk/
```
petergphillips, line 65: this project doesn't look like it has been migrated to use the generic service chart yet so you'll need to add this env var to ./templates/_envs.tpl as well

ldlharper: Done thanks.

helm_deploy/values-dev.yaml
ldlharper: Done thanks.

backend/tests/prisonerProfileService.test.ts
ldlharper: Done thanks.

petergphillips (approved):
LGTM

PR #1939, opened 2021-11-29
SLM-16: Add new homepage tile for Send Legal Mail

nathanrussell-moj-digital:


backend/controllers/homepage/homepage.ts
```diff
@@ -168,6 +168,13 @@ const getTasks = ({ activeCaseLoadId, locations, staffId, whereaboutsConfig, key
       href: bookAPrisonVisit.ui_url,
       enabled: () => bookAPrisonVisit.ui_url && userHasRoles(['PVB_STAFF']),
     },
+    {
```
sp-watson, line 171: We normally add a test in `homepageController.test.ts` for each new tile

helm_deploy/values-preprod.yaml
```diff
@@ -63,6 +63,7 @@ env:
   DISABLE_REQUEST_LIMITER: true
   PVB_URL: https://manage-a-prison-visit-preprod.hmpps.service.justice.gov.uk/
   CALCULATE_RELEASE_DATES_URL: https://calculate-release-dates-preprod.hmpps.service.justice.gov.uk
+  SEND_LEGAL_MAIL_URL: https://send-legal-mail-preprod.prison.service.justice.gov.uk
```
sp-watson, line 66: There isn't one of these for prod yet - just checking that is intentional?!

petergphillips: you might need to put a blank value in for `values-prod.yaml` otherwise the helm deployment could fail

backend/controllers/homepage/homepage.ts
nathanrussell-moj-digital: 👍 - OK, I've added a unit test in `homepageController.test.ts` and also a cypress test

helm_deploy/values-preprod.yaml
nathanrussell-moj-digital: It was semi intentional 😄 because I don't know what our prod url will be yet and we obviously don't want to run the risk of this tile being seen before we have released it properly.

In `homepage.ts` I have this:
```
    {
      id: 'send-legal-mail',
      heading: 'Check Rule 39 mail',
      description: 'Scan barcodes on mail from law firms and other approved senders.',
      href: sendLegalMail.url,
      enabled: () => sendLegalMail.url && userHasRoles(['SLM_SCAN_BARCODE', 'SLM_SECURITY_ANALYST']),
    },
```
which enables the tile if the url is truthy and the user has either role. 
In `config.ts` I have this:
```
  sendLegalMail: {
    url: process.env.SEND_LEGAL_MAIL_URL || '',
  },
```
My thinking was that if the prod helm config did not define the env var, the OR would kick in resulting in an empty string, which in turn is not truthy so the tile would not be enabled.

That was my thinking but being new to MoJ projects there is still a lot to learn about these things 😄 

@petergphillips - Is there a better way of doing this? Happy to do whatever is the norm 👍

nathanrussell-moj-digital: Actually, thinking about it, I think you are right @petergphillips in being explicit about these things. I'll add this to `values-prod.yaml`:
```
  SEND_LEGAL_MAIL_URL: ""
```

sp-watson (approved):

joelstobart-moj (approved):
LGTM

timharrison-moj (approved):
LGTM

PR #1947, opened 2021-12-06
NN-3797: fixing the text on the adjudications title and setting a case…

richardpopple:
…load for dev


helm_deploy/values-dev.yaml
```diff
@@ -59,7 +59,7 @@ env:
   PECS_URL: https://hmpps-book-secure-move-frontend-staging.apps.live-1.cloud-platform.service.justice.gov.uk
   COMPLEXITY_OF_NEED_URI: https://complexity-of-need-staging.hmpps.service.justice.gov.uk
   PRISONS_WITH_OFFENDERS_THAT_HAVE_COMPLEX_NEEDS: "MDI,LEI"
-  PRISONS_WITH_MANAGE_ADJUDICATIONS_ENABLED: ""
+  PRISONS_WITH_MANAGE_ADJUDICATIONS_ENABLED: "MDI"
```
sp-watson, line 62: If we do this, won't it work in Live?

sp-watson (approved):

PR #1948, opened 2021-12-06
NN-3653: new tile for adjudications

richardpopple:


natclamp-moj (dismissed):

helm_deploy/digital-prison-services/templates/_envs.tpl
```diff
@@ -274,4 +274,7 @@ env:
   - name: SEND_LEGAL_MAIL_URL
     value: {{ .Values.env.SEND_LEGAL_MAIL_URL | quote }}
 
+  - name: PPRISONS_WITH_MANAGE_ADJUDICATIONS_ENABLED
```
sp-watson, line 277: Double PP?

richardpopple: You are indeed correct!

sp-watson (approved):

PR #1953, opened 2021-12-09
DCS-1102 users without NMI caseload not prevented 

GurnankCheema:
from viewing Welcome service tile for prod env


louisenorris (approved):
LGTM

PR #1977, opened 2022-01-07
NN-2460: ✨ Add visit start time and remove status

petergphillips:
Screenshot from 2022-01-07 17-12-58

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,7 @@ jobs:
           command: npx cypress install
       - run:
           name: Run the node app.
-          command: npm run start-feature --env=cypress.env
+          command: npm run start-feature
           background: true
       - run:
           name: Run Cypress Integration tests
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ These tests use WireMock to stub the application's dependencies on the prisonApi
 ### Running the Cypress tests
 
 You need to fire up the wiremock server first:
-```docker-compose -f docker-compose-test.yaml up```
+```docker compose -f docker-compose-test.yaml up```
 
 This will give you useful feedback if the app is making requests that you haven't mocked out. You can see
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -39,7 +39,7 @@ These tests use WireMock to stub the application's dependencies on the prisonApi
 ### Running the Cypress tests
 
 You need to fire up the wiremock server first:
-```docker-compose -f docker-compose-test.yaml up```
+```docker compose -f docker-compose-test.yaml up```
```
petergphillips, line 42: the test compose file targets version `3.8` which is only compatible with `docker compose` as a cli plugin to docker

.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ jobs:
           command: npx cypress install
       - run:
           name: Run the node app.
-          command: npm run start-feature --env=cypress.env
+          command: npm run start-feature
```
petergphillips, line 66: we used to have groovy tests too, but since only cypress tests then don't need the `env` flag

views/prisonerProfile/prisonerVisits/prisonerVisits.njk
```diff
@@ -44,10 +44,6 @@
     {
       text: visit.relationship,
       classes: 'govuk-table__cell--last' if visit.isLast 
-    },
-    {
-      html: hideIfNotFirst(visit.status, visit.isFirst),
-      classes: 'govuk-table__cell--last' if visit.isLast
```
petergphillips, line 50: Remove the `status` column as it was incorrect.  Need to do more analysis to understand the visit data and then can add back in with better values e.g. only show `Attended` if visit is in the past and the visitor has been confirmed, show `Upcoming` for future visits etc.

andymarke (approved):
LGTM.

PR #1985, opened 2022-01-18
SDI-4: rename page

petergphillips:
- SDI-4: ✨ Add additional visit fields
- SDI-4: ✨ Rename page and add different no data message


backend/controllers/prisonerProfile/prisonerVisits.ts
```diff
@@ -100,6 +100,7 @@ export default ({ prisonApi, pageSize = 20 }) =>
         prisonerName: formatName(details.firstName, details.lastName),
         results,
         visitTypes: hasLength(visitTypes) && visitTypes.map((type) => ({ value: type.code, text: type.description })),
+        filterApplied: fromDate || toDate || visitType,
```
petergphillips, line 103: need a different message if there are no results when filters have been applied

andymarke (approved):
LGTM

PR #1987, opened 2022-01-19
SDI-4: ✨ Update visit status according to design

petergphillips:


backend/controllers/prisonerProfile/prisonerVisits.ts
```diff
@@ -7,6 +7,20 @@ export const VISIT_TYPES = [
...
 ]
+
+const calculateStatus = ({ cancelReasonDescription, completionStatusDescription, completionStatus, startTime }) => {
+  switch (completionStatus) {
+    case 'CANC':
+      return `Cancellation: ${cancelReasonDescription}`
+    case 'SCH': {
+      const start = moment(startTime, DATE_TIME_FORMAT_SPEC)
+      return start.isAfter(moment(), 'minute') ? 'Scheduled' : 'Not entered'
+    }
```
andymarke, line 18: So I guess that is saying if we have a scheduled visit in the past the status will be describe as `Not entered` - won't that many of historic visits will say this if the status in NOMIS is not updated after the visit?

andymarke (approved):
LGTM

backend/controllers/prisonerProfile/prisonerVisits.ts
petergphillips: agreed - from [SDI-4](https://dsdmoj.atlassian.net/browse/SDI-4):
> Where the visit is in the past, has no Completion outcome entered, and the ‘attended’ box is ticked, display visit status as ‘Not entered’ 
> Where the visit is in the past, has no Completion outcome entered, and the ‘attended’ box is unticked, display visit status as ‘Not entered’

PR #1993, opened 2022-01-21
SDI-4: ✨ Add status filter

petergphillips:


backend/controllers/prisonerProfile/prisonerVisits.ts
```diff
@@ -8,7 +8,26 @@ export const VISIT_TYPES = [
...
 ]
 
-const calculateStatus = ({ cancelReasonDescription, completionStatusDescription, completionStatus, startTime }) => {
+const calculateDateAndStatusFilter = (status: string, fromDate: string, toDate: string) => {
+  const fromAsDate = fromDate ? moment(fromDate, 'DD/MM/YYYY') : undefined
+  const toAsDate = toDate ? moment(toDate, 'DD/MM/YYYY') : undefined
+
+  const now = moment().startOf('day')
+  if (status === 'EXP') return { visitStatus: 'SCH', fromAsDate, toAsDate: toAsDate?.isBefore(now) ? toAsDate : now }
+  if (status === 'SCH') return { visitStatus: 'SCH', fromAsDate: fromAsDate?.isAfter(now) ? fromAsDate : now, toAsDate }
```
petergphillips, line 17: Similarly front end wants a `Scheduled` status to indicate visits in the past, so need to play around with the from and to dates to Prison API to cope

backend/controllers/prisonerProfile/prisonerVisits.ts
```diff
@@ -8,7 +8,26 @@ export const VISIT_TYPES = [
...
+  const splitStatus = status?.split('-')
```
petergphillips, line 18: also the status filter can contain all the cancellation reasons so need to split them out and search for them separately

backend/controllers/prisonerProfile/prisonerVisits.ts
```diff
@@ -101,6 +129,18 @@ export default ({ prisonApi, pageSize = 20 }) =>
           )
           .flat()
 
+      const statuses = cancellationReasons
+        .map((type) => ({ value: `CANC-${type.code}`, text: `Cancelled: ${type.description.toLowerCase()}` }))
```
petergphillips, line 133: cancellation reasons first

backend/controllers/prisonerProfile/prisonerVisits.ts
```diff
@@ -101,6 +129,18 @@ export default ({ prisonApi, pageSize = 20 }) =>
...
+        .concat(
+          completionReasons
```
petergphillips, line 135: then completion reasons (without `CANC` or `SCH`)

backend/controllers/prisonerProfile/prisonerVisits.ts
```diff
@@ -101,6 +129,18 @@ export default ({ prisonApi, pageSize = 20 }) =>
...
+            .filter((reason) => reason.code !== 'CANC' && reason.code !== 'SCH')
+            .map((type) => ({ value: type.code, text: capitalize(type.description) }))
+        )
+        .concat([
```
petergphillips, line 139: then the two past and future filters where the status isn't any of the other values

PR #1998, opened 2022-01-25
SDI-4: 🐛 Improve sorting and remove capitalisation of reference data

petergphillips:


backend/controllers/prisonerProfile/prisonerVisits.ts
```diff
@@ -8,6 +8,20 @@ export const VISIT_TYPES = [
   { value: 'OFFI', text: 'Official' },
 ]
 
+const sortByListSequenceThenDescription = (left, right): number => {
```
andymarke, line 11: just use Kotlin in the frontend then you can use the collection library :-)

andymarke (approved):
LGTM

PR #2006, opened 2022-01-28
ESWE-257-turn_feature-flag-off

monskapinga:


steverendell (commented):
I tihnk you need to set true in preprod as well so you can test it

steverendell (approved):

PR #2014, opened 2022-02-09
DCS-1381 change HDC card title and description

GurnankCheema:
HDC card now looks like this 
Screenshot 2022-02-09 at 15 28 34


andrewrlee (approved):
LGTM

PR #2017, opened 2022-02-14
HAAR-148 Add GP  and PF to allow lists

Mjwillis:


ushkarev (approved):

ushkarev:
a slight aside: i notice that the `dev` helm values file uses an older(?) `enable_whitelist` instead of `enable_allowlist` to control whether the ingress gets restricted. i imagine it does nothing and is harmless but thought i'd mention it anyway!

PR #2019, opened 2022-02-16
INC-88: Case notes behaviour prompts

ushkarev:
When adding a case note, specifically a "positive behaviour" or "negative behaviour" type, shows informative prompts to help officers with their decision-making.

behaviour-entry-prompt


backend/config.ts
```diff
@@ -19,6 +19,7 @@ export const app = {
   esweEnabled: process.env.ESWE_ENABLED === 'true',
   neurodiversityEnabledUsernames: process.env.NEURODIVERSITY_ENABLED_USERNAMES,
   disableRequestLimiter: process.env.DISABLE_REQUEST_LIMITER ? process.env.DISABLE_REQUEST_LIMITER === 'true' : false,
+  caseNotesBehaviourPrompts: process.env.CASE_NOTES_BEHAVIOUR_PROMPTS === 'true',
```
ushkarev, line 22: feature flag because this feature is not meant to go live before a particular date

Mjwillis (approved):
LGTM

PR #2032, opened 2022-02-25
ESWE-275_screen_format_and_additional_testing

monskapinga:


sass/_govuk-overrides.scss
```diff
@@ -92,4 +92,58 @@
...
+    @include govuk-media-query($from: tablet) {
+      width: 70%;
+    }
+  }
+
+  .govuk-summary-list__value:last-child {
+    @include govuk-media-query($from: tablet) {
+      width: 70%;
+    }
+  }
```
steverendell, line 148: Is this a copy of .prisoner-profile with just the widths changed? I think a better approach is to define 2 classes which just has the width defined, (with important! probably) and include them in the row.key.classes and row.value.classes fields

backend/services/esweService.ts
```diff
@@ -52,9 +52,12 @@ export const DEFAULT_GOALS = {
...
-  diversityAssessed: [],
-  diversitySelfDeclared: [],
-  diversitySupport: [],
+  divergenceAssessed: ['Alzheimers', 'Anxiety'],
+  assessmentDate: '20 February 2022',
+  divergenceSelfDeclared: ['Dyslexia', 'Autism'],
+  selfDeclaredDate: '10 February 2022',
+  divergenceSupport: ['Reading support', 'Communication'],
+  supportDate: '15 February 2022',
 }
```
steverendell, line 61: this looks like test data, maybe move it or remove?

steverendell (approved):

PR #2037, opened 2022-03-03
NN-3967: Cache contentful calls

sp-watson:


backend/routes.ts
```diff
@@ -78,6 +79,8 @@ const setup = ({
     next()
   })
 
+  const contentfulCache = cacheFactory(30) // 30 second TTL
```
sp-watson, line 82: We possibly should make this configurable?

richardpopple: Maybe, though I'm not sure we would gain that much? I guess if it was an env we would be able to pick it with out a new release - but we'd still have to to a redeploy for it to be picked up?

backend/utils/singleValueInMemoryCache.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  makeValueToCache(dataToCache: any): StoredCachedData {
+    return {
+      expiryTime: this.currentDateTime.add(this.ttlInSeconds, 'seconds'),
+      cachedData: dataToCache,
+    }
+  }
+}
+
+export const cacheFactory = (ttlInSeconds) => {
```
richardpopple, line 32: Unrelated - can you stick default here instead of at the bottom of the file?

backend/utils/singleValueInMemoryCache.ts
```diff
@@ -0,0 +1,54 @@
...
+  cachedValue = null
```
richardpopple, line 33: So I think I get what is going on here - but... should this 'cachedValue' only be in scope here? I.e. not declared at the top of the file. That way every time we call the cacheFactory to create a cache we will get a different local value - which I think means it no longer has to be single use. Like we could call 
const contentfulCache = cacheFactory(30) and then
const anotherServiceCache = cacheFactory(60)?

richardpopple (dismissed):

backend/utils/singleValueInMemoryCache.ts
sp-watson: Nope - this seems to be the only way to do it. Adding the `default export` at the bottom just appeased the linter - it isn't actually used

richardpopple (approved):

PR #2041, opened 2022-03-09
NN-3302: Fix add-appointment location list bug

natclamp-moj:
Currently, the list shown for available locations for an appointment is a list from the USER'S caseload. It needs to be a list of locations at the PRISONER'S caseload. I've swapped the user's agency id for the prisoner's. I've also sorted the typescript errors in the files involved.

### DIFF
--- a/backend/controllers/appointments/addAppointment.ts
+++ b/backend/controllers/appointments/addAppointment.ts
@@ -3,6 +3,12 @@ import { properCaseName, formatName } from '../../utils'
 import { buildDateTime, formatDate, DAY_MONTH_YEAR } from '../../../common/dateHelpers'
 import { repeatTypes, endRecurringEndingDate, validateComments } from '../../shared/appointmentConstants'
 
+export type PrePopulatedData = {
+  offenderEvents?: string[]
+  locationName?: string
+  locationEvents?: string[]
+}
+
 export const validateDate = (date, errors) => {
   const now = moment()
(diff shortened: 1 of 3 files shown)


backend/controllers/appointments/addAppointment.ts
```diff
@@ -131,22 +143,17 @@ export const addAppointmentFactory = (appointmentsService, existingEventsService
...
-      const prePopulatedData = {}
-      const { firstName, lastName, bookingId } = await prisonApi.getDetails(res.locals, offenderNo)
+      const prePopulatedData: PrePopulatedData = {}
+      const { firstName, lastName, bookingId, agencyId } = await prisonApi.getDetails(res.locals, offenderNo)
       const offenderName = `${properCaseName(lastName)}, ${properCaseName(firstName)}`
-      const { appointmentTypes, appointmentLocations } = await getAppointmentTypesAndLocations(
-        res.locals,
-        activeCaseLoadId
-      )
+      const { appointmentTypes, appointmentLocations } = await getAppointmentTypesAndLocations(res.locals, agencyId)
```
richardpopple, line 151: So we were passing the wrong thing in before - should have been agencyId?

natclamp-moj: yep - apparently it wasn't an issue before as users would only be booking prisoners into appointments and they'd both be in the same caseload. But now with video booking they often have to book appointments for prisoner's in other caseloads

PR #2051, opened 2022-03-17
SDI-30: ✨ Hide visit details if unsentenced

petergphillips:


backend/api/prisonApi.ts
```diff
@@ -323,7 +323,7 @@ export const prisonApiFactory = (client) => {
     get(context, `/api/bookings/${bookingId}/caseNotes/NEG/IEP_WARN/count?fromDate=${fromDate}&toDate=${toDate}`)
 
   const getPrisonerVisitBalances = (context, offenderNo) =>
-    get(context, `/api/bookings/offenderNo/${offenderNo}/visit/balances`)
+    get(context, `/api/bookings/offenderNo/${offenderNo}/visit/balances?allowNoContent=true`)
```
petergphillips, line 326: new parameter in prison api that returns no content (204) rather than not found (404) if there are no visit balances set

backend/controllers/prisonerProfile/prisonerQuickLook.ts
```diff
@@ -142,6 +142,23 @@ export default ({ prisonerProfileService, prisonApi, telemetry, offenderSearchAp
       return prisonerDetail?.indeterminateSentence ? 'Life sentence' : 'Not entered'
     }
 
+    const visitBalancesSection =
+      // @ts-expect-error ts-migrate(2339) FIXME: Property 'error' does not exist on type 'unknown'.
+      visitBalancesResponse.error || (visitBalances && Object.keys(visitBalances).length)
```
andymarke, line 147: so I guess we can't have remainingVo without Pvo being present, so just checking for no props works fine

andymarke (approved):
LGTM

backend/controllers/prisonerProfile/prisonerQuickLook.ts
petergphillips: agreed

PR #2062, opened 2022-03-25
SDI-129: ✨ Add in call to absence reasons v2

petergphillips:


backend/controllers/attendance/attendance.ts
```diff
@@ -62,8 +62,12 @@ export const attendanceFactory = (whereaboutsApi) => {
   }
 
   const getAbsenceReasons = async (context) => {
-    const absenceReasons = await whereaboutsApi.getAbsenceReasons(context)
+    const [absenceReasons, absenceReasonsV2] = await Promise.all([
+      await whereaboutsApi.getAbsenceReasons(context),
+      await whereaboutsApi.getAbsenceReasonsV2(context),
```
petergphillips, line 67: we call both whereabouts api absence reasons endpoints for now and then combine results

hyderkhanmoj: Is the idea to remove the v1 version in the future?

hyderkhanmoj (approved):
LGTM

backend/controllers/attendance/attendance.ts
petergphillips: Yes, well sort of, it is called from multiple places within DPS so having to do a bit at a time.  Idea is to gradually migrate all to new call, then copy `/v2` to the existing (now unused) endpoint, move everything back to the existing endpoint and then remove `/v2`.

PR #2067, opened 2022-03-28
SDI-129: ✨ Load existing sub reason into popup (second time lucky)

petergphillips:


integration-tests/integration/activity/activityList.spec.js
```diff
@@ -305,8 +305,8 @@ context('Activity list page', () => {
     })
     cy.get('[data-qa="other-message"').contains('Unacceptable - Incentive Level warning')
     cy.get('[data-qa="other-message"').parent().click({ multiple: true })
-    cy.get('[name="absentReason"]').contains('Unacceptable - Incentive Level warning')
-    cy.get('[name="absentSubReason"]').contains('Courses, programmes and interventions')
+    cy.get('[name="absentReason"]').find(':selected').contains('Unacceptable - Incentive Level warning')
```
petergphillips, line 308: this was checking the whole `select` rather than the selection option

PR #2073, opened 2022-03-29
INC-390: Copy change to 'Manage incentives' tile

xoen:
Removed reference to 'review dates' as we're removing the 'Days since
last review'/'Days on level' columns in 'Manage incentives'.


Mjwillis (approved):
LGTM

PR #2081, opened 2022-03-31
INC-293: call Incentives API to get list of IEP entries by prisoner

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-293


backend/api/incentivesApi.ts
```diff
@@ -12,10 +12,13 @@ export const incentivesApiFactory = (client) => {
 
   const getAgencyIepLevels = (context, agencyId) => get(context, `/iep/levels/${agencyId}`)
 
+  const getIepSummaryForBooking = (context, bookingId) => get(context, `/iep/reviews/booking/${bookingId}`)
```
neilmendum, line 15: Right now [IncentivesAPI calls](https://github.com/ministryofjustice/hmpps-incentives-api/blob/cc7e9b51670926cf490c8f90e32b771966740160/src/main/kotlin/uk/gov/justice/digital/hmpps/incentivesapi/service/PrisonApiService.kt#L29-L37) the PrisonAPI endpoint `POST:/bookings/iepSummary`. DPS used to call `GET:/booking/{bookingId}/iepSummary` which has the flag `withDetails` to decide if PrisonAPI includes `iepDetails`

Only place this was set to `false` was `prisonerQuickLook`. @Mjwillis do you think I should update IncentivesAPI to include this flag? Was initially thinking no, but now thinking that _prisonerQuickLook_ is a heavily used page and if possible would be good to reduce the JSON payloads it receives

Mjwillis: Well I be tempted to use https://incentives-api-dev.hmpps.service.justice.gov.uk/
Screenshot 2022-03-31 at 12 22 33
webjars/swagger-ui/index.html#/iep-level-resource/getCurrentIEPLevelForPrisoner ```POST /iep/reviews/bookings``` with either just one bookingId.  because I assume Quick look just needs to show the level

Mjwillis (commented):
Can you check if quick look needs anything other than the level name

Mjwillis (approved):
Approve but could change on next request

backend/api/incentivesApi.ts
neilmendum: That screen also uses other incentives data further down in the _Incentives and adjudications in last 3 months_ section - [iepSummary.daysSinceReview](https://github.com/ministryofjustice/digital-prison-services/blob/main/backend/controllers/prisonerProfile/prisonerQuickLook.ts#L133). Will change the IncentivesAPI endpoint to include the flag 👍

PR #2094, opened 2022-04-07
INC-470: call Incentives API to get IEP summaries when searching for prisoners

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-470


Mjwillis (approved):
LGTM - nice!

PR #2103, opened 2022-04-14
NN-4056: Update base docker image to fix security issue

sp-watson:


Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN CYPRESS_INSTALL_BINARY=0 npm ci --no-audit && \
 
 RUN npm prune --production
 
-FROM node:14-buster-slim
+FROM node:14.17-buster-slim
```
petergphillips, line 21: `14` will pick up `14.17` anyway, don't think this is required

sp-watson: Well it's best to be consistent, but also the security alert isn't going away by just re-deploying so i thought this might sort it.

I was in the process of checking the docker image itself, but I have had to address some other stuff (e.g App Insights), which is why I hadn't asked for this to be reviewed yet.

sp-watson: Well, I can't see why this vuln is reported here but not in e.g. `hmpps-restricted-patients-prod` docker image.

Doing a `dpkg -l | grep zlib` in all of these images gives:

zlib1g:amd64            1:1.2.11.dfsg-1        amd64        compression library - runtime

So hopefully by doing this update it will stop alerting 🤷

petergphillips: I can see from https://hub.docker.com/_/node that
```
14-buster-slim, 14.19-buster-slim, 14.19.1-buster-slim, fermium-buster-slim
```
are all the same docker image.  If you're wanting to pin to anything better to pin to the latest 14.19 instead.

sp-watson: Ah yes - ok will do

sp-watson:
On (much) further investigation the problem is that the `apt-get upgrade..` dockerfile command is been cached. We need to sort that out instead of updating the version.

It is very puzzling because a very similar Dockerfile in manage-key-workers doesn't have the caching problem?!?

PR #2131, opened 2022-05-05
NN-4019: Update node and npm

natclamp-moj:


.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   build:
     executor:
       name: hmpps/node
-      tag: 14.17-browsers
+      tag: 16.13-browsers
```
andrewrlee, line 10: meh.. looks like 16.15 is now the latest version

.circleci/config.yml
```diff
@@ -37,7 +37,7 @@ jobs:
   cypress_integration_tests:
     working_directory: ~/app
     docker:
-      - image: cimg/node:14.15.3-browsers
+      - image: cimg/node:16.13-browsers
       - image: bitnami/redis:5.0
```
andrewrlee, line 41: could potentially look at using the common executor [here]( https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/.circleci/config.yml#L101-L104), it might not "just work".

backend/setupWebSession.ts
```diff
@@ -13,13 +14,23 @@ export default () => {
...
     if (!enabled || !host) return null
 
-    const client = redis.createClient({
+    const client = createClient({
       host,
       port,
       password,
       tls: config.app.production ? {} : false,
+      socket: {
+        reconnectStrategy: (attempts: number) => {
```
andrewrlee, line 23: I think this is a new configuration option for v4 so this is unlikely to have an effect (though you will need it in v4).

It might be that the v3 version is at the top level rather than within `socket` but I think v3 has a default retry strategy which is reasonable anyway - `Math.min(retries * 50, 500)`.

andrewrlee: (In v4 the intent was to have the same default, there's just a [bug](https://github.com/redis/node-redis/issues/2095) that meant its not working - if I'd known that that was the default I probably would have gone with that rather than being fancy 🤷‍♂️ )

andrewrlee (approved):
lgtm

sp-watson (approved):

PR #2137, opened 2022-05-11
NN-4019: Add unhandled rejection handler to prevent pod crashes due to Node 16 upgrade

sp-watson:


backend/index.ts
```diff
@@ -31,6 +31,10 @@ import homepageController from './controllers/homepage/homepage'
 import deprecatedUrlPage from './controllers/deprecatedUrlPage'
 import requestLimiter from './middleware/requestLimiter'
 
+process.on('unhandledRejection', (reason, promise) => {
+  console.log('Unhandled Rejection at: ', promise, 'reason:', reason)
```
sp-watson, line 35: Not sure whether we should try to send this to App Insights, if we can - just getting it in there for now

natclamp-moj (approved):

petergphillips (approved):
LGTM

PR #2142, opened 2022-05-11
NN-4113: Remove batch of TS errors

natclamp-moj:


backend/controllers/cellMove/searchForCell.ts
```diff
@@ -3,7 +3,7 @@ import { putLastNameFirst, formatName, formatLocation } from '../../utils'
 
 import { userHasAccess, getNonAssocationsInEstablishment, renderLocationOptions, cellAttributes } from './cellMoveUtils'
 
-export default ({ oauthApi, prisonApi, whereaboutsApi }) =>
+export default ({ oauthApi, prisonApi, whereaboutsApi, logError }) =>
```
sp-watson, line 6: Maybe I am missing something, but why not just remove logError?

sp-watson (approved):

PR #2146, opened 2022-05-12
NN-4065 NN-4065 restricted patient access

timoj58:
changes to allow POM access to restricted patients in their case load.  

Note:  the cpress int tests fails specifically on a select (gov,uk) due to the following code not adding user in middleware

https://github.com/ministryofjustice/digital-prison-services/blob/14f70f32359238d5ba08293ad1386daff6d65086/backend/middleware/currentUser.ts#L30

Therefore the prisonerProfileContext is checking if the user is defined, if not, it will simply return the current context


Mjwillis (approved):
Ok, my only concern is are we now getting a token on every request or caching?

PR #2150, opened 2022-05-16
NN-4114: Caching of auth client credentials

sp-watson:


backend/api/systemOauthClient.ts
```diff
@@ -1,31 +1,87 @@
...
-      url: config.apis.oauth2.url,
-    }
-  ).makeTokenRequest(oauthRequest, 'PSH-client_credentials')
+  const oauthResult = await oauthClient.makeTokenRequest(oauthRequest, 'PSH-client_credentials')
 
   logger.debug(`Oauth request for grant type 'client_credentials', result status: successful`)
   return oauthResult
 }
 
+const debug = (operation: string, username: string) => {
```
sp-watson, line 49: We ought to remove this once we are confident it is working OK

backend/index.ts
```diff
@@ -48,6 +49,7 @@ app.set('view engine', 'njk')
 
 nunjucksSetup(app)
 phaseNameSetup(app, config)
+clientCredsSetup(getTokenStore(config), apis.oauthApi)
```
sp-watson, line 52: I had to add this setup code to facilitate testing

backend/api/systemOauthClient.ts
```diff
@@ -1,31 +1,87 @@
...
+  if (!enabled || !host) return null
+
+  const client = createClient({
+    host,
+    port,
+    password,
+    tls: config.app.production ? {} : false,
+  })
+
+  client.on('error', (e: Error) => logger.error('Redis client error', e))
```
steverendell, line 22: Doesnt the e parameter come first?

sp-watson: You mean in the `error` function parameters? Hmm.. yes.. which means the other 2 places in DPS are wrong as well. I'll fix them

Good spot!

timoj58 (approved):

PR #2151, opened 2022-05-17
NN-4065 include result of rp check for profile service edit mode

timoj58:


sp-watson (approved):
It's a bit difficult for me to review, given I didn't review the previous changes, but LGTM

PR #2158, opened 2022-05-19
ESWE-287_feature_flag_neurodivergence_indicator_repair

monskapinga:


backend/services/esweService.ts
```diff
@@ -10,8 +10,8 @@ import {
   LearnerLatestAssessment,
   LearnerProfile,
   LearnerEducation,
-  LearnerNeurodivergence,
 } from '../api/curious/types/Types'
+import { createFlaggedContent } from '../shared/neuroDivergenceHelper'
```
sp-watson, line 14: Not sure why this is here?!

backend/controllers/prisonerProfile/prisonerPersonal.ts
```diff
@@ -48,6 +50,19 @@ export default ({
     const treatmentCodes = treatmentTypes && treatmentTypes.map((type) => type.code).join()
     const healthCodes = healthTypes && healthTypes.map((type) => type.code).join()
 
+    const {
+      app: { neurodiversityEnabledUsernames, neurodiversityEnabledPrisons },
+    } = config
+
+    // TODO: Part of temporary measure to only allow restricted prisons access to neurodivergence data. If no prisons specified then assume all are allowed access.
+    const getNeurodivergenceSupportNeed = async () => {
+      let divergence = createFlaggedContent([])
```
sp-watson, line 59: Why flagged?

backend/shared/neuroDivergenceHelper.ts
```diff
@@ -0,0 +1,19 @@
+import { app } from '../config'
+
+export const canViewNeurodivergenceSupportData = (agencyId: string, enabledPrisons): boolean => {
```
sp-watson, line 3: Having got familiar with the code, it would seem that the `esweService` is the best place for this.

But it's not that important

backend/services/prisonerProfileService.ts
```diff
@@ -65,6 +69,19 @@ export default ({
 
     const systemContext = await systemOauthClient.getClientCredentialsTokens(username)
 
+    // TODO: Temporary measure to allow 3rd party MegaNexus time to implement caching and other techniques to their API so it can handle high volume of calls. To be refactored!
+    const getNeurodivergenceSupportNeed = async (agencyId) => {
+      if (canViewNeurodivergenceSupportData(agencyId, neurodiversityEnabledPrisons as string)) {
+        const [divergenceData] = await Promise.all(
+          [curiousApi.getLearnerNeurodivergence(systemContext, offenderNo)].map((apiCall) =>
```
sp-watson, line 76: As stated in the other PR - you ought to just use the standard promise in this situation.

But it isn't that important

backend/tests/prisonerPersonal.test.ts
```diff
@@ -2289,5 +2299,28 @@ describe('prisoner personal', () => {
...
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      esweService.getNeurodivergence = jest.fn()
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      curiousApi.getLearnerNeurodivergence.mockResolvedValue(neurodivergence)
+      await controller(req, res)
+      expect(res.render).toHaveBeenCalledWith(
+        'prisonerProfile/prisonerPersonal/prisonerPersonal.njk',
+        expect.objectContaining({ neurodivergence: undefined })
```
sp-watson, line 2322: Just to confirm - is `undefined` ok here?

backend/services/esweService.ts
monskapinga: _createFlaggedContent_ is now a shared resource, it's used in all methods calling the curiousApi in esweService.ts and is also used in prisonerPersonal.ts

backend/services/prisonerProfileService.ts
monskapinga: Point noted, will do in next PR

backend/controllers/prisonerProfile/prisonerPersonal.ts
sp-watson: Ok. I mean it's confusing because i don't see anywhere where the `enabled` part of it is ever referenced.

There is no point having flagged content that is shown without checking the flag?!

I guess this is something we can deal with later though

backend/tests/prisonerPersonal.test.ts
monskapinga: Yes - when prisoner is not in accelerated prison

sp-watson: I know, but in https://github.com/ministryofjustice/digital-prison-services/blob/ESWE-287_feature_flag_neurodivergence_indicator_repair/views/prisonerProfile/prisonerPersonal/partials/neurodiversities.njk#L3 it references a property in that value: `{% set neurodiversityInfo = neurodivergence.content %}` - does that not cause a rendering error?

sp-watson (approved):

PR #2173, opened 2022-05-31
Add a new DPS tile for the Legacy PVB

rossjones:
The legacy PVB is being re-animated for a short period of time whilst
its replacement is completed.  Previously PVB was only ever accessed via
its own service URL and therefore has never been used with DPS.

This PR adds a tile specifically for the legacy service so that we can
encourage users to access it via the preferred method.

Should resolve https://dsdmoj.atlassian.net/browse/VB-879

### DIFF
--- a/backend/tests/homepageController.test.ts
+++ b/backend/tests/homepageController.test.ts
@@ -21,6 +21,7 @@ describe('Homepage', () => {
     config.apis.manageAdjudications.ui_url = undefined
     config.apis.manageRestrictedPatients.ui_url = undefined
     config.apis.managePrisonVisits.ui_url = undefined
+    config.apis.legacyPrisonVisits.ui_url = undefined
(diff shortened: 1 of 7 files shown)


backend/tests/homepageController.test.ts
```diff
@@ -233,6 +234,27 @@ describe('Homepage', () => {
       )
     })
 
+    it.skip('should render home page with the legacy book a prison visit task', async () => {
```
petergphillips, line 237: did yo mean to put a `skip` here - what's the purpose of the test if it is skipped?

rossjones: Was just copying the existing PVB test and missed that. Will fix.

petergphillips (dismissed):
LGTM

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #2174, opened 2022-05-31
SDI-213: ♻️  Switch to new prisoners unaccounted for endpoint

petergphillips:
It still won't be speedy since there are issues with the Prison API endpoint.  Should be a bit better though as only making one call to get the data and also new 30 second timeout on the endpoint.

### DIFF
--- a/backend/api/prisonApi.ts
+++ b/backend/api/prisonApi.ts
@@ -274,9 +274,6 @@ export const prisonApiFactory = (client) => {
   const addSingleAppointment = (context, bookingId, body) =>
     post(context, `/api/bookings/${bookingId}/appointments`, body)
 
-  const getOffenderActivities = (context, { agencyId, date, period }) =>
-    get(context, `/api/schedules/${agencyId}/activities?date=${date}&timeSlot=${period}`)
-
   const getOffenderActivitiesOverDateRange = (context, { agencyId, fromDate, toDate, period }) =>
     get(
       context,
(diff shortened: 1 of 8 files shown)


Mjwillis (approved):
LGTM

backend/api/prisonApi.ts
```diff
@@ -274,9 +274,6 @@ export const prisonApiFactory = (client) => {
   const addSingleAppointment = (context, bookingId, body) =>
     post(context, `/api/bookings/${bookingId}/appointments`, body)
 
-  const getOffenderActivities = (context, { agencyId, date, period }) =>
```
petergphillips, line 277: removed - can now call whereabouts to get that info

backend/api/whereaboutsApi.ts
```diff
@@ -37,8 +37,10 @@ export const whereaboutsApiFactory = (client) => {
 
   const getAbsenceReasons = (context) => get(context, '/absence-reasons')
 
-  const getPrisonAttendance = (context, { agencyId, date, period }) =>
-    get(context, `/attendances/${agencyId}/attendance-for-scheduled-activities?date=${date}&period=${period}`)
+  const prisonersUnaccountedFor = (context, { agencyId, date, period }) =>
```
petergphillips, line 40: new endpoint with increased timeout as can take up to 30 seconds

Mjwillis (approved):
LGTM

PR #2175, opened 2022-06-01
NN-4190: Change content on RP tile

natclamp-moj:


backend/controllers/homepage/homepage.ts
```diff
@@ -219,7 +219,7 @@ const getTasks = ({ activeCaseLoadId, locations, staffId, whereaboutsConfig, key
       id: 'manage-restricted-patients',
       heading: 'Manage Restricted Patients',
       description:
-        'View all restricted patients in a secure hospital, move someone to a secure hospital, and remove someone from restricted patients.',
+        'View your restricted patients, move someone to a secure hospital, or remove someone from the Restricted Patients service.',
```
petergphillips, line 222: odd to see capitalisation of both Restricted and Patients, but then the heading is the same so 🤷.  I thought that our services should only be capitalised on the first letter.  Interesting also that the service heading is _Manage Restricted Patients_, but then it is referred to here as just _Restricted Patients_.

petergphillips (dismissed):
LGTM

backend/controllers/homepage/homepage.ts
natclamp-moj: I can pick it up with our content designer!

petergphillips (approved):
LGTM

PR #2187, opened 2022-06-10
SDI-226: ♻️  Switch to new suspended activities endpoint

petergphillips:


backend/controllers/attendance/attendanceStatistics.ts
```diff
@@ -386,12 +385,9 @@ export const attendanceStatisticsFactory = (oauthApi, prisonApi, whereaboutsApi)
...
     }
 
-    const offendersData = suspendedActivites.map((activity) => {
+    const offendersData = suspendedActivities.map((activity) => {
       const attendanceDetails = suspendedAttendances.attendances.find(
-        (attendance) =>
-          activity.bookingId === attendance.bookingId &&
-          moment(activity.startTime).format('YYYY-MM-DD') === attendance.eventDate &&
-          activity.eventId === attendance.eventId
+        (attendance) => activity.eventId === attendance.eventId
```
petergphillips, line 390: `eventId` is the primary key for an activity anyway, so no need to filter by extra criteria

backend/tests/attendanceStatistics.test.ts
```diff
@@ -662,18 +662,6 @@ describe('Attendance reason statistics', () => {
...
-          bookingId: 1133344,
-          offenderNo: 'G8977UK',
-          eventId: 5,
-          cellLocation: 'LEI-4',
-          startTime: '2019-10-10T14:00:00',
-          timeSlot: 'AM',
-          firstName: 'Offender',
-          lastName: 'Four',
-          comment: 'Cleaner',
-          suspended: false,
```
petergphillips, line 675: this was previously being filtered in the typescript, but no need now as Prison API only returns the suspended ones

backend/tests/attendanceStatistics.test.ts
```diff
@@ -696,7 +684,7 @@ describe('Attendance reason statistics', () => {
             comments: 'Did not ask nicely',
           },
           {
-            eventId: 4,
+            eventId: 5,
```
petergphillips, line 687: had to change the `eventId` here as the values are unique in the database

integration-tests/plugins/index.js
```diff
@@ -65,6 +65,7 @@ module.exports = (on) => {
         auth.stubUserMe(username, 12345, 'James Stuart', caseload),
         prisonApi.stubUserCaseloads(caseloads),
         tokenverification.stubVerifyToken(true),
+        keyworker.stubKeyworkerMigrated(),
```
petergphillips, line 68: this was annoying me as was causing all signins to display the sorry page instead of the home page

backend/controllers/attendance/attendanceStatistics.ts
petergphillips: The `EVENT_ID` comes from the `OFFENDER_COURSE_ATTENDANCES` table.  I think the confusion arises since in order to join to that table one needs to:
```
              JOIN OFFENDER_COURSE_ATTENDANCES OCA ON OCA.OFFENDER_BOOK_ID = OPP.OFFENDER_BOOK_ID AND
                    OCA.EVENT_DATE = CS.SCHEDULE_DATE AND
                    OCA.CRS_SCH_ID = CS.CRS_SCH_ID
```
so if we had a `courseScheduleId` instead we would need to use all three fields.

I have confirmed this by calling the `activities-by-date-range` endpoint for a period of a week and checking that each activity has a different event id.

andymarke (approved):
Lgtm

PR #2190, opened 2022-06-13
HAAR-500 Add prevent double click on add appointments, and case notes

Mjwillis:
Signed-off-by: mjwillis <michael.willis@digital.justice.gov.uk>


brightonsbox (approved):
LGTM

PR #2200, opened 2022-06-28
NN-3244: Change offender to prisoner in cell moves url

natclamp-moj:
Swapping out the term 'offender' for 'prisoner' in cell moves URL. I've deliberately left the variable name(s) as I think there's keys on pieces of data that use offenderNo and I don't want that to get more confusing.


sp-watson (approved):
It looks ok but the problem is that we cannot easily check if these urls are used elsewhere.

i guess that means we need to do some serious testing?!

PR #2204, opened 2022-06-30
NN-3240: Cell moves CSRA data changes

natclamp-moj:


backend/controllers/cellMove/viewCellSharingAssessmentDetails.ts
```diff
@@ -26,7 +26,7 @@ export default ({ prisonApi, logError }) =>
         prisonerName: putLastNameFirst(firstName, lastName),
         cellLocation: (assignedLivingUnit && assignedLivingUnit.description) || 'Not entered',
         location: (location && location.description) || 'Not entered',
-        level: mostRecentAssessment && mostRecentAssessment.classification,
+        level: mostRecentAssessment && translateCsra(mostRecentAssessment.classificationCode),
```
sp-watson, line 29: This is not fixing the problem is it?

backend/controllers/cellMove/cellMoveUtils.ts
```diff
@@ -50,11 +51,17 @@ export const cellAttributes = [
   { text: 'Multiple occupancy', value: 'MO' },
 ]
 
+export const translateCsra = (csraClassificationCode: string): string => {
+  if (csraClassificationCode === 'not entered') return 'not entered'
```
sp-watson, line 55: Not sure why you are using a hard-coded string for the input? Why not just use `if (!csraClassification)`?

Oh... it's because of the current code that calls it. What a mess!

backend/controllers/cellMove/considerRisks.ts
```diff
@@ -56,8 +57,8 @@ export default ({ prisonApi, raiseAnalyticsEvent }) => {
       const currentOffenderWithOccupants = [currentOffenderDetails, ...currentOccupantsDetails]
 
       const offendersCsraValues = currentOffenderWithOccupants
-        .filter((currentOccupant) => currentOccupant.csra)
-        .map((currentOccupant) => currentOccupant.csra)
+        .filter((currentOccupant) => translateCsra(currentOccupant.csraClassificationCode))
```
sp-watson, line 60: What's the point of this `filter`?

backend/controllers/cellMove/selectCell.ts
```diff
@@ -74,7 +79,7 @@ const getCellOccupants = async (res, { prisonApi, activeCaseLoadId, cells, nonAs
             nonAssociations.nonAssociations &&
             nonAssociations.nonAssociations.find((na) => na.offenderNonAssociation.offenderNo === occupant.offenderNo)
         ),
-        csra: csraInfo && csraInfo.classification,
+        csra: (csraInfo && translateCsra(csraInfo.classificationCode)) || 'Not entered',
```
sp-watson, line 82: Not sure this fixes the problem?

backend/controllers/cellMove/cellMoveUtils.ts
natclamp-moj: I could change the code where it's used to not pass in a default, and then return 'not entered' here when undefined is passed in?

backend/controllers/cellMove/selectCell.ts
natclamp-moj: See other response, I assume it's the same idea

backend/controllers/cellMove/viewCellSharingAssessmentDetails.ts
sp-watson: Sort out later - it needs us to look at Prison API

sp-watson (approved):

PR #2210, opened 2022-07-06
DCS-1689: ✨ Add endpoint to store a referring system

connormaglynn:
This PR supports the capability to allow services to navigate users to DPS, and configure DPS to provide dynamic navigation back to the original service.

To do this, I have created a new endpoint called **/save-backlink** which takes in the following query parameters:
- service (required) - identifies the service which maps to a pre-registered hostname and default options.
- returnResource (required) - allows the service to dynamically return the user to different resources, depending on the journey.
- toResource (required) - where to redirect the user within DPS after storing backlink information. Allowing services to dynamically send users to different parts of DPS depending on the journey.
- backLinkText (optional) - Optional text that will be displayed within the DPS profile, each service has default text that is displayed.

I've also migrated the **View most recent search** functionality over to use this more general approach. The visible difference in this functionality due to this migration are:
- The prisoner links on the prisoner search page read as **/save-backlink?...** instead of directly to the prisoner-page (though the users are redirected to the same place as before)


backend/controllers/backLink.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (!service || !returnResource || !toResource) throw new Error('Required query parameters missing')
+
+  const userBackLink = registeredServices.find((e) => e.name === service)
+  if (!userBackLink) throw new Error(`Could not find service: [${service}]`)
+
+  req.session.userBackLink = {
+    url: userBackLink.hostname + returnResource,
+    text: backLinkText || userBackLink.defaultBackLinkText,
+  }
+  res.redirect(toResource)
```
andrewrlee, line 36: We can't allow the client to specify the location for arbitrary redirects

integration-tests/integration/search/prisonerSearch.cy.js
```diff
@@ -269,7 +269,11 @@ context('Prisoner search', () => {
           })
       })
 
-      cy.get('[data-test="recent-search-link"]').should('have.attr', 'href', searchUrl)
+      cy.get('[data-test="back-link"]').should(
+        'have.attr',
+        'href',
+        '/prisoner-search?keywords=Saunders&location=MDI&alerts[]=XA,URCU'
```
andrewrlee, line 275: this `[]` syntax is fine and brought in from the body parser I think

backend/controllers/backLink.ts
connormaglynn: Prefixed with the application url, so all redirects are scoped to be within DPS. For some reason, this is what I assume **res.redirect** did anyways 😂 Probably because I've only ever used it for that! 😁

backend/controllers/backLink.ts
```diff
@@ -0,0 +1,37 @@
...
+    name: 'digital-prison-services',
+    hostname: '',
+    defaultBackLinkText: 'View most recent search',
+  },
+  {
+    name: 'welcome-people-into-prison',
+    hostname: sanitizeUrl(config.apis.welcomePeopleIntoPrison.url),
+    defaultBackLinkText: 'Back to Welcome people into prison',
+  },
+] as Array<RegisteredService>
```
andrewrlee, line 22: you don't need to do a type assertion here with the `as`, if you want to be explicit about the type you can: `export const registeredBackLinkServices:  RegisteredService[] = ...`

backend/controllers/backLink.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const saveBackLink = (registeredServices: Array<RegisteredService>) => async (req, res) => {
```
andrewrlee, line 24: you could default this to the production `registeredBackLinkServices`, save exporting it to just inject it back in later and its a reasonable default.

backend/controllers/backLink.ts
```diff
@@ -0,0 +1,37 @@
...
+  const { service, returnResource, toResource, backLinkText } = req.query
+
+  if (!service || !returnResource || !toResource) throw new Error('Required query parameters missing')
```
andrewrlee, line 27: `toResource` => `redirectPath` ? 
`returnResource` => `returnPath` ?

backend/controllers/backLink.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const userBackLink = registeredServices.find((e) => e.name === service)
+  if (!userBackLink) throw new Error(`Could not find service: [${service}]`)
+
+  req.session.userBackLink = {
+    url: userBackLink.hostname + returnResource,
```
andrewrlee, line 33: if you specify `service` = `digital-prison-services` then you can `returnResource = https://my-evil-website.com`

backend/tests/backLink.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      backLinkText: 'backLinkText',
+    }
+    await controller(req, res)
+    expect(req.session.userBackLink).toEqual({
+      text: 'backLinkText',
+      url: `${wpipUrl}/returnResource`,
+    })
+  })
+
+  it.each([[...registeredServices]])('should save text as default when not provided', async (registeredService) => {
```
andrewrlee, line 81: Might be best to not `it.each` these - is there any reason why any service would act differently?

backend/tests/backLink.test.ts
```diff
@@ -0,0 +1,93 @@
+import { RegisteredService, saveBackLink } from '../controllers/backLink'
+
+const wpipUrl = 'https://wpip'
+const dpsUrl = 'https://dps'
+jest.mock('../config', () => ({
+  app: { url: dpsUrl },
+  apis: { welcomePeopleIntoPrison: { url: wpipUrl } },
```
andrewrlee, line 7: meh.. as you're injecting test versions of the the services then this shouldn't be needed?

views/prisonerSearch/partials/prisonerSearchList.njk
```diff
@@ -1,7 +1,7 @@
 {% set rows = [] %}
 {% for prisoner in results %}
   {% set prisonerImage = '<img src="/app/images/' + prisoner.offenderNo + '/data" alt="Image of ' + prisoner.name + '" class="prisoner-search__results-list__image" />' %}
-  {% set prisonerLink = '<a href="/prisoner/' + prisoner.offenderNo + '" class="govuk-link">' + prisoner.name + '</a>'%}
+  {% set prisonerLink = '<a href="/save-backlink?service=digital-prison-services&returnResource=/prisoner-search%3Fkeywords%3D' + formValues.keywords + '%26location%3D' + formValues.location +  '%26alerts[]%3D' + formValues.alerts + '&toResource=/prisoner/' + prisoner.offenderNo + '" class="govuk-link">' + prisoner.name + '</a>'%}
```
andrewrlee, line 4: is this trying to capture the current url? Would it be safe to pass `req.url` through here instead?

backend/controllers/backLink.ts
connormaglynn: Set it to the apps url, I did this to start with then got rid - far too much confidence in my assumption that res.redirect was all internal 😂

views/prisonerSearch/partials/prisonerSearchList.njk
connormaglynn: Looked into it and seems like since part of the URL is query params, it's gets confused unless the **returnResource**  has encoded values. Not sure if there's an out of the box query param encoder, or if we'd have to write a custom one. Think sticking with building up the url is best suited here?

andrewrlee: theres this kind of thing, and decoder
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

andrewrlee: It's obviously quite fragile having to keep this url up to date with the application and bugs might be quite subtle a new filter is added but this isn't updated

andrewrlee (approved):
I still have some reservations around this. 

I still think it might be better to have a central place for all the individual entry points into DPS in here, rather than having them scattered across different service repos.

The interface to use this (urls you need to build up) are quite complex (4 params) and will need to be maintained across multiple of repos. 

Links can break quite easily, if DPS gets updated but external systems don't, whereas if you had a simpler interface and DPS controlled redirect logic, you could have tests for all the entry points.   

Allowing clients to specify redirect locations is quite risky and as we've seen its easy for subtle security issues to sneak in, I _think_ this is ok...  

Maybe we can try it out and see how it goes, it looks like we have a few use cases for this now so hopefully will be able to learn from that. 
LGTM

connormaglynn:
Just noticed there's a lot more query params in the search page. Going to go with @andrewrlee suggestions of passing in the URL to the template 👍 Put PR to draft for time being

andymarke:
> Allowing clients to specify redirect locations is quite risky and as we've seen its easy for subtle security issues to sneak in, I _think_ this is safe now..

@andrewrlee won't you **have** to do this some how? Wouldn't clients want to return back to a dynamic url, e.g go back to list of prisoners in a specific 
 prison in say MPC or whatever?

andymarke:
It feels like like we should be able to simplify this to something like 

`/prisoners/A1234AK?returnUrl=https://pathfinder.service.hmpps.gov.uk/mynominals?region=london`

Then you can just build up an allowlist of domains when needed:

config: {
 redirectDomainsAllowList: ['pathfinder.service.hmpps.gov.uk', 'someotherdomain']
}

Then the middleware is looking for the special `returnUrl` or some such thing.

michalnawrockidj (approved):
LGTM

connormaglynn:
> It feels like like we should be able to simplify this to something like
> 
> `/prisoners/A1234AK?returnUrl=https://pathfinder.service.hmpps.gov.uk/mynominals?region=london`
> 
> Then you can just build up an allowlist of domains when needed:
> 
> config: { redirectDomainsAllowList: ['pathfinder.service.hmpps.gov.uk', 'someotherdomain'] }
> 
> Then the middleware is looking for the special `returnUrl` or some such thing.

Hi @andymarke, thanks for having a nosey at this! 😁 I see what you mean with the simplification; I think this would be a good idea if we were always landing on the prisoner profile. For our journeys, though, we want to send the user to other parts of DPS, more specifically, to add a case note on the initial landing. 

Merged in the changes - I think it's good enough to go; it might need some refinement as we add other journeys into it. I know Cell Moves within DPS uses backlinks as buttons on the page, so some sort of display configuration may be required 🤔 We'll cross that bridge when we get to it, though! 🌉

andymarke:
Sounds good to me

PR #2218, opened 2022-07-12
NN-4324: Removing ts error comments

natclamp-moj:


backend/api/oauthApi.test.ts
```diff
@@ -5,7 +5,7 @@ const clientId = 'clientId'
 const url = 'http://localhost/'
 const clientSecret = 'clientSecret'
 
-const client = {}
+const client = {} as any
```
sp-watson, line 8: This isn't great but I suppose it's better than not having a type... does the client not actually have an object?

backend/utils.ts
```diff
@@ -2,31 +2,37 @@ import moment from 'moment'
...
-export const formatTimestampToDateTime = (timestamp, format = 'DD/MM/YYYY - HH:mm') =>
-  timestamp && moment(timestamp).format(format)
+export const formatTimestampToDateTime = (
+  timestamp: moment.MomentInput,
+  format = 'DD/MM/YYYY - HH:mm'
+): moment.MomentInput => timestamp && moment(timestamp).format(format)
 
-export const distinct = (data) =>
-  data.reduce((accumulator, current) => (accumulator.includes(current) ? accumulator : [...accumulator, current]), [])
+export const distinct = (data: never[]): unknown =>
```
sp-watson, line 28: never?

backend/utils.ts
```diff
@@ -317,7 +328,7 @@ export const getWith404AsNull = async (apiCall) =>
       })
   )
 
-export const stringWithAbbreviationsProcessor = (string) => {
+export const stringWithAbbreviationsProcessor = (string: string): string => {
```
sp-watson, line 331: This is good work - something that none of us would ever get round to!

sp-watson (approved):

PR #2219, opened 2022-07-12
NN-4324: Removing TS error comments

natclamp-moj:


backend/controllers/cellMove/cellMoveConfirmation.ts
```diff
@@ -1,6 +1,6 @@
 import { formatName, putLastNameFirst } from '../../utils'
 
-export default ({ prisonApi }) =>
+export default ({ prisonApi, logError }) =>
```
sp-watson, line 3: You seem to use underscore elsewhere when the var isn't used (which i like) - why not use it here (and other ones in this PR like this)?

sp-watson: i have been told it causes warnings in this case, so we'll leave it as is

PR #2220, opened 2022-07-13
INC-473: exclude incentives tile appearing if user has no locations or with Central admin caseload

neilmendum:
https://dsdmoj.atlassian.net/browse/INC-473


helm_deploy/values-prod.yaml
```diff
@@ -35,6 +35,7 @@ env:
   USE_OF_FORCE_PRISONS: "ACI,AGI,ASI,AYI,BAI,BCI,BFI,BHI,BLI,BMI,BNI,BRI,BSI,BWI,BXI,BZI,CDI,CFI,CLI,CWI,DAI,DGI,DHI,DMI,DNI,DTI,DWI,EEI,EHI,ESI,EWI,EXI,EYI,FBI,FDI,FHI,FKI,FMI,FNI,FSI,FWI,GHI,GMI,GNI,GTI,HBI,HCI,HDI,HEI,HHI,HII,HLI,HMI,HOI,HPI,HVI,ISI,IWI,KMI,KVI,LCI,LEI,LFI,LGI,LHI,LII,LLI,LNI,LPI,LTI,LWI,LYI,MDI,MHI,MRI,MSI,MTI,NHI,NLI,NMI,NSI,NWI,ONI,OWI,PBI,PDI,PFI,PNI,PRI,PVI,RCI,RHI,RNI,RSI,SDI,SFI,SHI,SKI,SLI,SNI,SPI,STI,SUI,SWI,TCI,TSI,UKI,UPI,VEI,WCI,WDI,WEI,WHI,WII,WLI,WMI,WRI,WSI,WTI,WWI"
   INCENTIVES_URL: https://incentives-ui.hmpps.service.justice.gov.uk
   INCENTIVES_API_ENDPOINT_URL: https://incentives-api.hmpps.service.justice.gov.uk
+  INCENTIVES_EXCLUDED_CASELOADS: "CADM_I"
```
neilmendum, line 38: Found this in [nomis-user-roles-api](https://github.com/ministryofjustice/nomis-user-roles-api/blob/2e5af2b10b1dc3b5263fc8f904158270b76867ac/src/main/resources/db/migration/V1_15__CASELOADS_DATA.sql#L26)

xoen (approved):

backend/controllers/homepage/homepage.ts
```diff
@@ -60,7 +60,17 @@ const getTasks = ({ activeCaseLoadId, locations, staffId, whereaboutsConfig, key
       description: 'See prisoner incentive information by residential location and view incentive data visualisations.',
       href: incentives.ui_url,
       roles: null,
-      enabled: () => incentives.ui_url,
+      enabled: () =>
+        incentives.ui_url &&
+        !(
+          incentives.excludedCaseloads.split(',').includes(activeCaseLoadId) ||
+          userHasRoles([
```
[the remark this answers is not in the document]
Mjwillis: You don't need this - use locations size e.g. https://github.com/ministryofjustice/digital-prison-services/pull/2220/files#diff-f8a9492c5a04c0a52f307b83958be45fd23f684f6dafeb35e41a3df8aeaaa675L108

`enabled: () =>  incentives.ui_url && Boolean(locations?.length > 0),`

backend/tests/homepageController.test.ts
```diff
@@ -552,6 +553,96 @@ describe('Homepage', () => {
...
+          expect.objectContaining({
+            tasks: [],
+          })
+        )
+
+        // set this back to empty list
+        config.apis.incentives.excludedCaseloads = ''
+      })
+
+      it('should not render home page with the Incentives tile if user is an LSA', async () => {
```
Mjwillis, line 586: You don't need to do this. You wouldn't want to exclude users because of these roles.  These roles can be assigned to any user and then they wouldn't get the tile.  The caseload is good enough and this size of the location dropdown above

Mjwillis (changes requested):
Remove role check and check location size instead

Mjwillis (approved):
LGTM

PR #2221, opened 2022-07-13
NN-4324: Removing expect errors

natclamp-moj:


backend/controllers/cellMove/cellNotAvailable.ts
```diff
@@ -1,6 +1,6 @@
 import { putLastNameFirst } from '../../utils'
 
-export default ({ prisonApi }) =>
+export default ({ prisonApi, logError }) =>
```
sp-watson, line 3: Same as per other PR (https://github.com/ministryofjustice/digital-prison-services/pull/2219) - use underscore?

sp-watson: i have been told it causes warnings in this case, so we'll leave it as is

PR #2229, opened 2022-07-20
NN-4346

timoj58:
oauthApi no longer calls the auth server, to get role, it decodes token instead.  In future, this could be moved out of  this client, but that would involve to many changes for one PR

This PR already contains a lot of changes namely

-  decode token
-  remove all reference to promise in unit tests and code 
-  refactor all int tests, including separate context where tests need to change permissions mid flow


Mjwillis (approved):
LGTM

PR #2240, opened 2022-08-15
SDI-311 Search feature now sticky and env driven

andymarke:
Still off by default

### DIFF
--- a/integration-tests/integration/search/prisonerSearch.cy.js
+++ b/integration-tests/integration/search/prisonerSearch.cy.js
@@ -80,7 +80,7 @@ context('Prisoner search', () => {
           data: [inmate1, inmate2],
         })
         cy.task('stubGetIepSummaryForBookingIds', [inmate1Iep, inmate2Iep])
-        cy.visit(`/prisoner-search`)
+        cy.visit(`/prisoner-search?feature=legacy`)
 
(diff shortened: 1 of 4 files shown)


integration-tests/integration/search/prisonerSearch.cy.js
```diff
@@ -80,7 +80,7 @@ context('Prisoner search', () => {
           data: [inmate1, inmate2],
         })
         cy.task('stubGetIepSummaryForBookingIds', [inmate1Iep, inmate2Iep])
-        cy.visit(`/prisoner-search`)
+        cy.visit(`/prisoner-search?feature=legacy`)
```
andymarke, line 84: for the cypress test force the testing of both feature routes - this will be remove once this is live

views/prisonerSearch/partials/prisonerSearchForm.njk
```diff
@@ -58,6 +58,8 @@
       text: "Search",
       type: "submit"
     }) }}
+
+    <input type="hidden" name="feature" value={{formValues.feature}} />
   </div>
```
andymarke, line 63: so that feature switch override works after a submit

lbennett-moj (approved):
LGTM

PR #2251, opened 2022-09-30
INC-790 Record an incentive level review

ushkarev:
This is largely content changes in order to encourage people to record an incentive level review even when the prisoner's level is not being changed. No functionality is being changed.

NB: DPS already showed the date of next review as 1 year after the last incentive level entry. This logic is incorrect but is not being changed yet as part of this PR – it will in future be calculated by the [incentives api](https://github.com/ministryofjustice/hmpps-incentives-api).


backend/tests/prisonerQuickLook.test.ts
```diff
@@ -451,7 +458,7 @@ describe('prisoner profile quick look', () => {
               details: [
                 { label: 'Incentive level warnings', value: 0 },
                 { label: 'Incentive encouragements', value: 0 },
-                { label: 'Last incentive level review', value: '0 days ago' },
+                { label: 'Date of next review', html: 'Unable to show this detail' },
```
ushkarev, line 461: I think _pretending_ that the last incentive review was "0 days ago" when data cannot be loaded was wrong. In my opinion we should state when we cannot show correct information.

xoen: I think you're right about being more explicit (a user may see "0 days ago" and wrongly conclude the prisoner just had a reviewed).

This is when somehow the request to the API failed right?
I wonder if a better wording would be "Unable to retrieve this detail" - but maybe it's too low level?

views/prisonerProfile/prisonerIncentiveLevelDetails.njk
```diff
@@ -42,15 +42,15 @@
     <div class="horizontal-information govuk-!-margin-bottom-3">
         <div class="horizontal-information__item">
             <h3 class="govuk-heading-s govuk-!-margin-bottom-1">Current incentive level</h3>
-            <p class="govuk-body" data-test="incident-time">{{ currentIepLevel }}</p>
+            <p class="govuk-body" data-test="current-iep">{{ currentIepLevel }}</p>
         </div>
         <div class="horizontal-information__item">
             <h3 class="govuk-heading-s govuk-!-margin-bottom-1">Time since review</h3>
-            <p class="govuk-body" data-test="incident-location">{{ daysOnIepLevel }}</p>
+            <p class="govuk-body" data-test="days-since-review">{{ daysOnIepLevel }}</p>
```
xoen: I know they may often be the same but `data-test` attribute suggests this "days since last review" but the value name suggests is "days on level" 🤔

Assuming the value is correct maybe change to `data-test="days-on-level"`?

backend/controllers/prisonerProfile/prisonerQuickLook.ts
```diff
@@ -46,6 +46,21 @@ const createFinanceLink = (offenderNo, path, value) =>
     formatCurrency(value || 0)
   }</a>`
 
+const getNextIepReviewDate = (iepSummary: { iepTime?: string }): string | undefined => {
+  const lastReviewDate = iepSummary?.iepTime
+  if (!lastReviewDate) return undefined
+  // TODO: nextReviewDate will come from incentivesApi in future
+  const nextReviewDate = moment(lastReviewDate, 'YYYY-MM-DD HH:mm').add(1, 'years')
```
xoen, line 53: You could have a similar logic to try and get it from the API response - and if not present sum 1 year like you already do.

This would mean that when the API will start to have this field DPS will automatically use it (and then this logic could be removed at a later stage)

something like:

```JavaScript
let nextReviewDate = iepSummary?.nextReviewDate
if (!nextReviewDate) {
  // fallback while Incentives API doesn't return this field
  nextReviewDate = moment(lastReviewDate, 'YYYY-MM-DD HH:mm').add(1, 'years')
}
```

it's just an idea :)

views/prisonerProfile/prisonerIncentiveLevelDetails.njk
ushkarev: the only point of this commit was to highlight that the existing label was meaningless (and unused). in a later commit, i change it to something different

xoen (dismissed):
I added a couple of comments/ideas but overall this looks great ☺️

backend/controllers/prisonerProfile/prisonerQuickLook.ts
ushkarev: `let nextReviewDate = iepSummary?.nextReviewDate` that's exactly what's coming "soon". once the incentives api starts providing this (even if somewhat incorrectly at first)

views/prisonerProfile/prisonerIncentiveLevelDetails.njk
xoen: Oh I see, sorry I was reviewing this commit-by-commit and I commented on something that changed. I see these labels are now correct :)

backend/tests/prisonerQuickLook.test.ts
ushkarev: "Unable to show this detail" is the text already used all over this page and i think it's best to keep it consistent; i don't wanna make the call to rewrite this piece of content everywhere

backend/controllers/prisonerProfile/prisonerQuickLook.ts
xoen: But you could have this in place right now and the UI would start displaying that `nextReviewDate` coming from the Incentives API as soon as this is in the response :)

xoen: 1. UI checks if this field is in the API response, if not calculate `nextReviewDate` as it currently does
2. Incentives API adds `nextReviewDate` to the IEP summary response (as in 1 year to begin with). UI starts to automatically display value from Incentives API
3. clean up UI code to remove fallback code (not needed anymore once this field is in Incentives API)

That was my thinking behind this idea.

backend/tests/prisonerQuickLook.test.ts
xoen: Oh there are other parts of DPS which use the exact same wording. Cool in that case consistency FTW :)

backend/controllers/prisonerProfile/prisonerIncentiveLevelDetails.ts
```diff
@@ -74,7 +74,9 @@ export default ({ prisonApi, incentivesApi, oauthApi }) =>
         }
       })
 
-      const nextReviewDate = moment(iepSummary.iepTime, 'YYYY-MM-DD HH:mm').add(1, 'years').format('D MMMM YYYY')
+      // TODO: nextReviewDate will come from incentivesApi in future
```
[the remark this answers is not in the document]
ushkarev: we don't have one yet, but i can put it in *this* thread once we do!

backend/tests/prisonerQuickLook.test.ts
```diff
@@ -548,12 +555,37 @@ describe('prisoner profile quick look', () => {
...
         )
       })
+
+      it('should show when next incentive review is overdue when in the past', async () => {
+        await controller(req, res)
+
+        const context = res.render.mock.calls.at(-1)?.[1]
+        const nextReviewDate = context?.incentives?.details?.at(-1)?.html
+        expect(nextReviewDate).toContain('15 August 2018')
+        expect(nextReviewDate).toContain('516 days overdue')
```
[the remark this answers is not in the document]
ushkarev: good idea; ✅

neilmendum (approved):
LGTM

backend/controllers/prisonerProfile/prisonerIncentiveLevelDetails.ts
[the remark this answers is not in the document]
ushkarev: Relevant stories:
• incentives-api tweak to calculate the date [INC-815](https://dsdmoj.atlassian.net/browse/INC-815)
• DPS changes to remove all these "TODO" calculations [INC-798](https://dsdmoj.atlassian.net/browse/INC-798) [INC-799](https://dsdmoj.atlassian.net/browse/INC-799) [INC-800](https://dsdmoj.atlassian.net/browse/INC-800)

PR #2253, opened 2022-09-30
INC-797: Display confirmation when incentive level is recorded

ushkarev:


backend/controllers/prisonerProfile/prisonerChangeIncentiveLevelDetails.ts
```diff
@@ -40,6 +42,38 @@ export default ({ prisonApi, incentivesApi }) => {
...
 
+  const renderConfirmation = async (req, res) => {
+    const { offenderNo } = req.params
+
+    try {
+      const prisonerDetails = await prisonApi.getDetails(res.locals, offenderNo, true)
+      const { agencyId, bookingId, firstName, lastName, assignedLivingUnit } = prisonerDetails
+      const locationId: string | undefined = assignedLivingUnit?.description
+
+      const iepSummary = await incentivesApi.getIepSummaryForBooking(res.locals, bookingId, false)
```
ushkarev, line 53: once incentives api starts to serve data from own DB, `iepSummary` could be used to get `locationId`. until then, it seems like `prisonApi.getDetails` needs to be used with "full info".

neilmendum (approved):
LGTM

PR #2255, opened 2022-10-03
NN-4417 generic service upgrade with existing ingress

timoj58:
As part of this the old redirect ingress is being removed.  If you look at the deleted .njk users were warned this will be removed, so one year later seems fine.

There is a very small amount of traffic registering on prometheus, generally all together, and often in the early hours, i do bot believe this is human interaction.


helm_deploy/digital-prison-services/values.yaml
```diff
@@ -1,12 +1,69 @@
...
+  image:
+    repository: quay.io/hmpps/digital-prison-services
+    tag: latest
+    pullPolicy: IfNotPresent
+    port: 3000
+
+  ingress:
+    enabled: true
+    v1_2_enabled: true
+    v0_47_enabled: false
```
neilmendum, line 15: For no downtime I thought we would want to enable both ingress controllers? Then once `v1_2` is serving traffic we can disable `v0_47`?

helm_deploy/digital-prison-services/templates/ingress-redirect.yaml
```diff
@@ -1,32 +0,0 @@
-{{- if .Values.ingress.enabled -}}
```
neilmendum, line 1: Probably worth mentioning we checked in Prometheus and this was no longer used?

helm_deploy/values-prod.yaml
```diff
@@ -1,178 +1,170 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
+generic-service:
+  replicaCount: 4
```
neilmendum, line 4: Not a huge one but typically people set this in `values.yaml` then need to only override in `values-dev.yaml`

helm_deploy/values-dev.yaml
```diff
@@ -1,80 +1,73 @@
...
+  replicaCount: 2
 
-replicaCount: 2
+  ingress:
+   host: digital-dev.prison.service.justice.gov.uk 
+
+  allowlist: false 
+
+  env:
+   PRISON_STAFF_HUB_UI_URL: digital-dev.prison.service.justice.gov.uk
```
neilmendum, line 12: I think this should be prefixed with `https://`? `_envs.tpl` had
```
- name: PRISON_STAFF_HUB_UI_URL
  value: "https://{{ .Values.ingress.host }}/"
```

neilmendum: Ditto for `values-preprod` and `values-prod`

helm_deploy/digital-prison-services/values.yaml
```diff
@@ -1,12 +1,69 @@
...
 generic-prometheus-alerts:
   targetApplication: digital-prison-services
 
 ingress:
   # live-1 is blue, live is green
   contextColour: green
   annotations:
-    external-dns.alpha.kubernetes.io/aws-weight: "100"
+    external-dns.alpha.kubernetes.io/aws-weight: "0"
     nginx.ingress.kubernetes.io/custom-http-errors: "418"
```
neilmendum, line 69: I don't think we need this because it's provided by the [generic-service](https://github.com/ministryofjustice/hmpps-helm-charts/blob/ead8a533ef3683c8ffda83b9080723f8ed99dfe1/charts/generic-service/values.yaml#L52)?

helm_deploy/digital-prison-services/values.yaml
```diff
@@ -1,12 +1,69 @@
...
+  
 generic-prometheus-alerts:
   targetApplication: digital-prison-services
 
 ingress:
   # live-1 is blue, live is green
   contextColour: green
   annotations:
-    external-dns.alpha.kubernetes.io/aws-weight: "100"
+    external-dns.alpha.kubernetes.io/aws-weight: "0"
```
neilmendum, line 68: I think this is also provided by the [generic-service](https://github.com/ministryofjustice/hmpps-helm-charts/blob/ead8a533ef3683c8ffda83b9080723f8ed99dfe1/charts/generic-service/templates/ingress.yaml#L20)?

helm_deploy/digital-prison-services/values.yaml
```diff
@@ -1,12 +1,69 @@
...
+    pullPolicy: IfNotPresent
+    port: 3000
+
+  ingress:
+    enabled: true
+    v1_2_enabled: true
+    v0_47_enabled: false
+    host: app-hostname.local    # override per environment
+    tlsSecretName: digital-prison-services-cert
+    contextColour: green
```
neilmendum, line 18: I think this has been dropped from the `generic-service` now that the _live-1_ to _live_ migration has been completed?

helm_deploy/digital-prison-services/values.yaml
neilmendum: As discussed we are not enabling `v0_47` because DPS has `ingress.yaml` for the current ingress, and we're enabling `v1_2`

helm_deploy/digital-prison-services/values.yaml
neilmendum: Same as https://github.com/ministryofjustice/digital-prison-services/pull/2255#discussion_r986689908

neilmendum (approved):
LGTM

PR #2260, opened 2022-10-04
INC-814: Improve incentives api types

ushkarev:
• Adds parameter and return types to incentives api object 
• Fixes some types in OAuth client api (e.g. POST methods _do_ in fact expect a body param)
• Fixes test mocks (a few used incorrect types, but test _logic_ all remains unchanged)
• Only functional code change is: IEP summary details aren't loaded on the incentive recording page (because they're unused)

### DIFF
--- a/backend/api/oauthEnabledClient.ts
+++ b/backend/api/oauthEnabledClient.ts
@@ -23,7 +23,7 @@ type OauthApiClientOptions = {
   timeout?: number
 }
 
-type ClientResponse<T = never> = {
+type ClientResponse<T = any> = {
   body: T
   req: ClientRequest
 } & superagent.Response
@@ -118,7 +118,7 @@ export class OauthApiClient {
    * @param body
    * @returns A Promise which resolves to the superagent result object, or the superagent error object if it is rejected
    */
-  post = <T>(context: ClientContext, path: string, body: never): Promise<ClientResponse<T>> => {
+  post = <T>(context: ClientContext, path: string, body: any): Promise<ClientResponse<T>> => {
     return superagent
       .post(this.createUrl(path))
       .send(body)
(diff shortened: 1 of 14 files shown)


backend/tests/prisonerChangeIncentiveLevelDetails.test.ts
```diff
@@ -9,40 +10,43 @@ jest.mock('../raiseAnalyticsEvent', () => ({
 
 describe('Prisoner change incentive level details', () => {
   const offenderNo = 'ABC123'
-  const bookingId = '123'
+  const bookingId = 123
```
ushkarev, line 13: many tests had this field as a string, but it's always a number

backend/api/oauthEnabledClient.ts
```diff
@@ -118,7 +118,7 @@ export class OauthApiClient {
    * @param body
    * @returns A Promise which resolves to the superagent result object, or the superagent error object if it is rejected
    */
-  post = <T>(context: ClientContext, path: string, body: never): Promise<ClientResponse<T>> => {
+  post = <T>(context: ClientContext, path: string, body: any): Promise<ClientResponse<T>> => {
```
ushkarev, line 121: you normally _do_ want to supply a body to a POST endpoint

xoen (approved):

backend/api/oauthEnabledClient.ts
ushkarev: it's ok to not include a body when `any` is asked for. but you cannot pass _any_ body if it requires `never`!

you can still call `client.post(context, '/…', undefined)` on new version cos an `undefined` counts as an `any`.

xoen: Ah OK, so the original was wrong but this makes `POST` requests' body optional which is what you want 👌

PR #2269, opened 2022-10-14
INC-860 Fix DPS incentive level details page when history includes system generated reviews

Mjwillis:
For unmatched records we now visually get "not entered" which we can iterate on by stops the screen crashing


ushkarev:
might be worth having a unit test to show the page loading should the prison api return 404 on a user lookup?

PR #2272, opened 2022-10-24
SDI-311 Turn on elatsic search for establishment search

andymarke:
Turn on in dev and preprod


petergphillips (approved):
LGTM

petergphillips:
Probably worth defining the value in the prod configuration too with value of `false`

PR #2276, opened 2022-10-31
Stop calling incentives-api when prisoner search finds no results

ushkarev:
…because it's started to enforce validation (submitted booking ids cannot be an empty list)


andymarke (approved):
LGTM

PR #2281, opened 2022-11-02
NN-4458 clear backlink from session

jamesgrant-moj:
**Bug**
Backlink that is stored in the session in the local establishment search isn't cleared out when searching using global search. 

**Solution**
Clear the saved backlink when using global search.


HamHazJustice (approved):
good

natclamp-moj (approved):
LGTM - looks like the PR title needs changing to "NN-4458..."

jamesgrant-moj:
Thanks for reviewing both. I updated the PR title.

PR #2309, opened 2022-12-15
SDI-509 House block appended with `dash`

andymarke:
This ensures search for block `C` doesn't bring back prisoner in block `Cswap`


backend/tests/prisonerSearch.test.ts
```diff
@@ -131,18 +131,18 @@ describe('Prisoner search', () => {
       expect(offenderSearchApi.establishmentSearch).toHaveBeenCalledWith(expect.anything(), 'BXI', {})
     })
 
-    it('should request with parsed location when supplied in query', async () => {
-      req.query.location = 'BXI-B-1-0'
```
andymarke, line 135: changed to realistic data - we never search for cells only blocks

petergphillips (approved):
LGTM

PR #2316, opened 2022-12-22
INC-1002: Update content for case note entry interruption when an incentive level should be recorded

ushkarev:


neilmendum (approved):
LGTM

PR #2322, opened 2023-01-04
NN-4109 - restyle UI controls on visitor page

HamHazJustice:


timoj58:
this branch needs to have the latest main changes merged into it

views/prisonerProfile/prisonerVisits/partials/prisonerVisitsSearchForm.njk
```diff
@@ -15,7 +15,7 @@
     <div class="horizontal-form govuk-!-margin-bottom-3">
       {{ datePicker({
         id: 'fromDate',
-        label: 'Date from',
+        label: 'From',
```
natclamp-moj, line 18: Is changing this label critical to correcting the styling? We shouldn't really change things like this in the design without a really good reason/if a designer says it needs doing

hamidhazrati: For the sake of consistency with other date from/to labels, then I think we can.
This label appears 6/7 times in 3 other 'njk' files
Perhaps we can create a ticket in backlog to address those to?
@SimonMitchellMOJ  any thoughts ?

HamHazJustice: SimonC agrees with you , so reverting those back.

natclamp-moj (approved):

PR #2332, opened 2023-01-11
NN-4741: add maintenance page which can be enabled for views which invoke whereabouts-api to modify data

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-4741


backend/routes.ts
```diff
@@ -268,48 +321,8 @@ const setup = ({
 
   router.use('/global-search', globalSearchRouter({ offenderSearchApi, oauthApi, logError }))
 
-  router.get('/manage-prisoner-whereabouts/select-location', selectActivityLocation({ prisonApi }).index)
-  router.post('/manage-prisoner-whereabouts/select-location', selectActivityLocation({ prisonApi }).post)
```
neilmendum, line 272: Moved these up to be with other `/manage-prisoner-whereabouts/` routes

backend/tests/whereaboutsMaintenanceMode.test.ts
```diff
@@ -0,0 +1,122 @@
+import supertest from 'supertest'
+import express from 'express'
+import nunjucksSetup from '../utils/nunjucksSetup'
+import routes from '../routes'
+
+describe('Whereabouts maintenance mode flag is true', () => {
```
neilmendum, line 6: Would have liked to also write a Cypress test for this but because the flag is set using `process.env.WHEREABOUTS_MAINTENANCE_MODE` I couldn't think of an easy way to toggle it. Open to ideas!

There's good Cypress test coverage of whereabouts and appointmentDetails and they still pass which is comforting

andrewrlee: yes, it's a pain. Having some kind of feature toggle system that could modify stuff at runtime would allow testing this. 

In projects I've worked with we generally move tests over to test the new feature and leave the old feature untested - which isn't ideal and doesn't make sense in this case

views/maintenancePage.njk
```diff
@@ -0,0 +1,35 @@
...
+
+  <div class="homepage-content">
+    <div class="govuk-width-container">
+
+      {% set mainClasses = "govuk-main-wrapper--l" %}
+      <div class="govuk-grid-row govuk-!-margin-top-7 govuk-!-margin-bottom-7">
+        <div class="govuk-grid-column-two-thirds">
+          <h1 class="govuk-heading-l">This service is currently unavailable</h1>
+          <p class="govuk-body">
+            You'll be able to use the service on DPS again from [6pm on Thursday, 1 December 2022].
```
neilmendum, line 25: Will update the date & time in another PR

natclamp-moj (approved):

PR #2335, opened 2023-01-16
SDI-284 Remove unnecessary profile links for Restricted Patients

mikehalmamoj:


backend/api/restrictedPatientApi.ts
```diff
@@ -2,22 +2,25 @@ export const processError = (error) => {
...
 }
 
 export const restrictedPatientApiFactory = (client) => {
   const processResponse = (agencyIds) => (response) => {
     const { agencyId, active } = response.body.supportingPrison
-    return agencyIds.includes(agencyId) && active
+    return {
+      isCaseLoadRestrictedPatient: agencyIds.includes(agencyId) && active,
+      hospital: response.body.hospitalLocation,
+    }
```
mikehalmamoj, line 14: Also need the hospital location

backend/controllers/prisonerProfile/prisonerAlerts.ts
```diff
@@ -14,7 +14,7 @@ export default ({
   async (req, res) => {
     const { offenderNo } = req.params
 
-    const { context, overrideAccess } = await getContext({
```
mikehalmamoj, line 17: It's not as simple as allowing POMs access to DPS anymore - we need other details from Restricted Patients

backend/controllers/prisonerProfile/prisonerProfileContext.ts
```diff
@@ -1,26 +1,33 @@
...
-  if (userRoles.map((userRole) => userRole.roleCode).includes('POM')) {
-    const isRestrictedPatient = await restrictedPatientApi.isCaseLoadRestrictedPatient(
-      res.locals,
-      offenderNo,
-      allCaseloads.map((caseload) => caseload.caseLoadId)
-    )
-    if (isRestrictedPatient) {
-      const context = await systemOauthClient.getClientCredentialsTokens(username)
-      return { context, overrideAccess: true }
+  const restrictedPatient = await restrictedPatientApi.getRestrictedPatientDetails(
```
petergphillips, line 13: doesn't this mean we're now calling the restricted patient api for every single prisoner? That's quite a big increase in the number of calls. Previously we only used to call it if the user was a POM.

petergphillips (approved):
LGTM

PR #2336, opened 2023-01-16
NN-4774: Security/dependence update/fixes

natclamp-moj:
- Add code to allowlist to ignore unnecessary json5 complaints
- update dependencies
- fix any problems arising from dep updates (react class names in snapshots etc)


neilmendum (approved):
LGTM

PR #2349, opened 2023-01-24
SDI-592 Turn off datacompliance in dev

mikehalmamoj:
The datacompliance API is broken in dev - not sure why, think it's related to secret rotation - but for now we need to turn off datacompliance requests from DPS to get DPS working again.


steverendell (approved):
lgtm

PR #2391, opened 2023-02-27
INC-1103: Removed 'Date of last review' from Incentive details page

xoen:
This date is currently not accurate. This is the date of the last Incentive record which is not necessarily an Incentive review (it could be a transfer for example).


ushkarev (approved):
Is the backend code remaining in case this element is reinstated?

PR #2403, opened 2023-03-17
MDWFP-92 Adding warrant folder button.

ldlharper:
Adding a link into the warrant folder service. This will only appear if users have the digital warrant folder role, which has not been given to any real world users.


yasinmustafa (approved):
LGTM

natclamp-moj (approved):

natclamp-moj (approved):

PR #2412, opened 2023-03-29
New homepage tile for Social video calls application (Phonehub)

rj-adams:


helm_deploy/values-dev.yaml
```diff
@@ -55,6 +55,7 @@ generic-service:
    MANAGE_RESTRICTED_PATIENTS_URL: https://manage-restricted-patients-dev.hmpps.service.justice.gov.uk/
    MANAGE_PRISON_VISITS_URL: https://manage-prison-visits-dev.prison.service.justice.gov.uk/
    LEGACY_PRISON_VISITS_URL: https://prison-visits-booking-staff-dev.apps.live.cloud-platform.service.justice.gov.uk/
+   SECURE_SOCIAL_VIDEO_CALLS_URL: https://auth.dev.prisonvideo.com/accounts/login/oauth/redirect/
```
neilmendum, line 58: Do you want to include URLs for other environments (assuming there are ones!)?

neilmendum (approved):
LGTM

Made a comment about links for other envs but happy to merge as it is if you are 😄

helm_deploy/values-dev.yaml
rj-adams: I don't have URLS for the other environments yet - but would like to push this into Dev so the PM and others can see how it operates

PR #2421, opened 2023-04-13
SDIT-738: ✨ Add in lazy loading and caching of offender images

petergphillips:


backend/controllers/images.ts
```diff
@@ -38,6 +38,7 @@ export const imageFactory = (prisonApi) => {
       prisonApi
         .getPrisonerImage(res.locals, offenderNo, fullSizeImage)
         .then((data) => {
+          res.set('Cache-control', 'private, max-age=86400')
```
petergphillips, line 41: set to be private so that only cached in browsers and not in intermediate servers.  86400 is one day

views/globalSearch/partials/globalSearchList.njk
```diff
@@ -1,6 +1,6 @@
 {% set rows = [] %}
 {% for prisoner in results %}
-  {% set prisonerImage = '<img src="/app/images/' + prisoner.offenderNo + '/data" alt="Image of ' + prisoner.name + '" class="prisoner-search__results-list__image" />' %}
+  {% set prisonerImage = '<img src="/app/images/' + prisoner.offenderNo + '/data" alt="Image of ' + prisoner.name + '" class="prisoner-search__results-list__image" loading="lazy" />' %}
```
petergphillips, line 3: just added lazy loading to two endpoints at present to try out

andymarke: (which BTW is not supported in Safari but I guess that doesn't matter)

andymarke (approved):
LGTM

PR #2431, opened 2023-04-25
INC-1238: Show Manage incentives tile to appropriate user groups

ushkarev:
The tile shows for:
• regular users in a prison – would have at least one "location" _irrespective_ of their roles
• central admins – won’t have locations but will have `MAINTAIN_INCENTIVE_LEVELS` role

LSAs would need to use their normal account. An LSA’s special account would have the active case load of `CADM_I` but no locations.

NB: first commit is just to rearrange test cases; the tests themselves are unchanged

### DIFF
--- a/backend/config.ts
+++ b/backend/config.ts
@@ -118,7 +118,6 @@ export const apis = {
   },
   incentives: {
     ui_url: process.env.INCENTIVES_URL,
-    excludedCaseloads: process.env.INCENTIVES_EXCLUDED_CASELOADS || '',
   },
   calculateReleaseDates: {
     ui_url: process.env.CALCULATE_RELEASE_DATES_URL,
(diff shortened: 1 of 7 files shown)


backend/config.ts
```diff
@@ -118,7 +118,6 @@ export const apis = {
   },
   incentives: {
     ui_url: process.env.INCENTIVES_URL,
-    excludedCaseloads: process.env.INCENTIVES_EXCLUDED_CASELOADS || '',
```
ushkarev, line 121: i don't think we need to have any roles configurable per-environment

integration-tests/integration/homepage/homepage.cy.js
```diff
@@ -221,10 +221,7 @@ context('Homepage', () => {
       page.incentives().tile().should('exist')
       page.incentives().title().contains('Manage incentives')
       page.incentives().link().should('exist')
-      page
-        .incentives()
-        .description()
-        .contains('See prisoner incentive information by residential location and view incentive data visualisations.')
+      page.incentives().description().should('exist')
```
ushkarev, line 224: the content here doesn't really matter as there's no logic to make it vary

Mjwillis (approved):
LGTM

PR #2449, opened 2023-05-11
NN-5081 upgrade typescript and front end

timoj58:


tsconfig.json
```diff
@@ -16,7 +16,8 @@
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "ignoreDeprecations": "5.0"
```
p-nik, line 20: do we need to ignoreDeprecations if tests were upgraded or fixed?

p-nik (approved):

tsconfig.json
timoj58: yes, tests are only upgraded due to accordion changes in govuk

PR #2451, opened 2023-05-12
CDPS-141- redirects - e2e tests

jamesgrant-moj:


backend/config.ts
```diff
@@ -4,13 +4,13 @@ const toNumber = (value: string | undefined): number | undefined => {
...
 export const app = {
-  port: process.env.PORT || 3002,
+  port: process.env.PORT || 3000,
   production: process.env.NODE_ENV === 'production',
   disableWebpack: process.env.DISABLE_WEBPACK === 'true',
   licencesUrl: process.env.LICENCES_URL || 'http://localhost:3003/',
   mailTo: process.env.MAIL_TO || 'feedback@digital.justice.gov.uk',
   tokenRefreshThresholdSeconds: toNumber(process.env.TOKEN_REFRESH_THRESHOLD_SECONDS) || 60,
-  url: process.env.PRISON_STAFF_HUB_UI_URL || `http://localhost:${process.env.PORT || 3002}/`,
+  url: process.env.PRISON_STAFF_HUB_UI_URL || `http://localhost:${process.env.PORT || 3000}/`,
```
brightonsbox, line 13: I can't remember if it was me that changed this - can you remember why the default port is changing here?

brightonsbox: Maybe would be good to change it back if we don't have a particular reason for change - who knows if this breaks other people's set ups etc!

jamesgrant-moj: Thanks for taking a look. Good spot Jon, I reverted it back

helm_deploy/values-dev.yaml
```diff
@@ -65,3 +65,6 @@ generic-service:
    KEYWORKER_MAINTENANCE_MODE: false
    GET_SOMEONE_READY_FOR_WORK_URL: https://get-ready-for-work-dev.hmpps.service.justice.gov.uk
    WARRANT_FOLDER_URL: https://manage-a-warrant-folder-dev.hmpps.service.justice.gov.uk/
+   PRISONER_PROFILE_REDIRECT_URL: http://prisoner-dev.digital.prison.service.justice.gov.uk
+   PRISONER_PROFILE_REDIRECT_ENABLED: true
```
brightonsbox, line 69: Can we switch these to (in all environments):

```
   PRISONER_PROFILE_REDIRECT_ENABLED: false
   PRISONER_PROFILE_REDIRECT_ENABLED_PRISONS: LEI
```

We can test using the `DPS_APPLICATION_DEVELOPER` role before switching ENABLED: true

integration-tests/integration/globalsearch/globalSearch.cy.js
```diff
@@ -35,41 +35,45 @@ context('Global search', () => {
...
-      expect(JSON.parse(requests[0].body)).to.deep.equal({
-        prisonerIdentifier: 'A1234BC',
-        gender: 'ALL',
-        location: 'ALL',
-        includeAliases: true,
-      })
+  context('should call global search', () => {
+    beforeEach(() => {
+      cy.visit('/global-search')
+      cy.task('stubGlobalSearch')
```
brightonsbox, line 41: do we need to stub before visiting? seems like this would be a bit of a race condition?

backend/config.ts
```diff
@@ -4,7 +4,7 @@ const toNumber = (value: string | undefined): number | undefined => {
 }
 
 export const app = {
-  port: process.env.PORT || 3002,
+  port: process.env.PORT || 3000,
```
brightonsbox, line 7: Can we switch this back to 3002?

brightonsbox (approved):
LGTM

PR #2490, opened 2023-07-04
MAP-36: Look up non-asociations by offender no instead of booking ID

jimbali:
Previously, non-associations for cell moves were looked up via the current booking number for the offender, meaning that potentially relevant non-associations from previous bookings were not shown. It has been changed to look them up via the offender number, which should include non-associations from previous bookings too.

I have hidden this behind a feature flag, as we in the MAP team are new to the project and I don't want to block other teams from releasing to production while we test in dev.


neilmendum:
tagging @ministryofjustice/hmpps-non-associations for their visibility

danbenton-mojdt (approved):

Thource (approved):

PR #2496, opened 2023-07-10
MAP-36: Remove non-associations legacy mode feature flag

jimbali:
This feature has been released to production now so we no longer need the feature flag.


backend/api/prisonApi.ts
```diff
@@ -433,11 +433,7 @@ export const prisonApiFactory = (client) => {
   const getScheduledEventsForNextWeek = (context, bookingId) =>
     get(context, `/api/bookings/${bookingId}/events/nextWeek`)
 
-  const getNonAssociations = (context, { bookingId, offenderNo }) => {
-    if (config.app.nonAssociationsLegacyMode) {
-      return get(context, `/api/bookings/${bookingId}/non-association-details`)
-    }
-
+  const getNonAssociations = (context, offenderNo) => {
     return get(context, `/api/offenders/${offenderNo}/non-association-details`)
```
GurnankCheema, line 437: why have we started to use a different prisonApi endpoint? 
Has there been a requirement to get non-associations for all bookings rather than latest booking?

jimbali: yep that was the objective of this ticket

GurnankCheema (approved):
LGTM

PR #2502, opened 2023-07-14
MAP-158: Filter non-associations based on current location

jimbali:
Previously we were filtering non-associations based on the location from the non-association record. This creates problems if the non-association was created against a previous booking, as the last location will be `OUTSIDE`. Now we check the current location of each prisoner and compare them instead.


GurnankCheema (approved):
LGTM

PR #2506, opened 2023-07-20
MAP-163 new view for reception areas

GurnankCheema:
new view and controller created to enable movement of prisoners to reception. The page is not accessible in the current flow but can be accessed by directly entering the url  (/cell-move/select-reception).
The /cell-move/select-cell page was used as a basis for the new page and the exact content isn't known yet as it is work in progress. Bare-bones tests added.

### DIFF
--- a/backend/controllers/cellMove/selectReception.ts
+++ b/backend/controllers/cellMove/selectReception.ts
@@ -49,7 +49,7 @@ const getCellOccupants = async (res, { prisonApi, activeCaseLoadId, cells, nonAs
     .map(
       (offenderNumber) =>
         assessmentsGroupedByOffenderNo[offenderNumber]
-          .filter((assessment) => assessment?.assessmentDescription.includes('CSR'))
+          .filter((assessment) => assessment?.assessmentDescription?.includes('CSR'))
           .sort(sortByLatestAssessmentDateDesc)[0]
     )
     .filter(Boolean)
@@ -72,9 +72,7 @@ const getCellOccupants = async (res, { prisonApi, activeCaseLoadId, cells, nonAs
         viewOffenderDetails: `/prisoner/${occupant.offenderNo}/cell-move/prisoner-details`,
         alerts: alertFlagLabels.filter((label) => label.alertCodes.some((code) => alertCodes.includes(code))),
         nonAssociation: Boolean(
-          nonAssociations &&
-            nonAssociations.nonAssociations &&
-            nonAssociations.nonAssociations.find((na) => na.offenderNonAssociation.offenderNo === occupant.offenderNo)
+          nonAssociations?.nonAssociations?.find((na) => na.offenderNonAssociation.offenderNo === occupant.offenderNo)
         ),
         csra: (csraInfo && translateCsra(csraInfo.classificationCode)) || 'Not entered',
         csraDetailsUrl: `/prisoner/${occupant.offenderNo}/cell-move/cell-sharing-risk-assessment-details`,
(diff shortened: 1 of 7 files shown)


backend/controllers/cellMove/selectReception.ts
```diff
@@ -0,0 +1,236 @@
...
+  const occupantAssessments = await prisonApi.getCsraAssessments(res.locals, occupantOffenderNos)
+  const assessmentsGroupedByOffenderNo = occupantAssessments ? groupBy(occupantAssessments, 'offenderNo') : []
+
+  const cellSharingAssessments = Object.keys(assessmentsGroupedByOffenderNo)
+    .map(
+      (offenderNumber) =>
+        assessmentsGroupedByOffenderNo[offenderNumber]
+          .filter(
+            (assessment) =>
+              assessment && assessment.assessmentDescription && assessment.assessmentDescription.includes('CSR')
```
steverendell, line 54: you can just use assessment?.assessmentDescription?.includes('CSR') here ?
(And there are several other similar situations in this file)

views/cellMove/selectReception.njk
```diff
@@ -0,0 +1,161 @@
...
+          <th scope="col" class="govuk-table__header">Relevant alerts</th>
+          <th scope="col" class="govuk-table__header"><span class="govuk-visually-hidden">Select cell</span></th>
+        </tr>
+      </thead>
+      <tbody class="govuk-table__body">
+        {% for cell in cells %}
+          {% if cell.occupants.length > 0 %}
+              {% for occupier in cell.occupants %}
+                <tr class="govuk-table__row">
+                  {{ makeRow(cell, loop.index, cell.occupants.length === loop.index, occupier) }}
```
steverendell, line 132: there is a loop.last boolean you could use

backend/controllers/cellMove/selectReception.ts
GurnankCheema: agreed

views/cellMove/selectReception.njk
GurnankCheema: at this stage, the display of reception on this page was merely a proof of concept. So it's not really necessary to change the logic too much when compared with the nunjucks template this page was cloned from.

steverendell: not sure what you mean, i was just suggesting changing 'cell.occupants.length === loop.index' to the simpler but identical 'loop.last'

GurnankCheema: oh I see

backend/controllers/cellMove/selectReception.ts
```diff
@@ -0,0 +1,233 @@
...
+    offenderNumbers: occupantOffenderNos,
+  })
+  const occupantAssessments = await prisonApi.getCsraAssessments(res.locals, occupantOffenderNos)
+  const assessmentsGroupedByOffenderNo = occupantAssessments ? groupBy(occupantAssessments, 'offenderNo') : []
+
+  const cellSharingAssessments = Object.keys(assessmentsGroupedByOffenderNo)
+    .map(
+      (offenderNumber) =>
+        assessmentsGroupedByOffenderNo[offenderNumber]
+          .filter((assessment) => assessment?.assessmentDescription.includes('CSR'))
```
steverendell, line 52: this is not quite the same as previously, i.e. do you know that assessmentDescription cant be null?
Also I was suggesting you could make a similar change e.g. at lines 75, 92, 104

PR #2513, opened 2023-07-28
MAP-202: Fix cell moves bug with non-associations in other establishments

jimbali:
We recently updated the cell moves code to pull the details of each non-association via the prison API. This uses the currently logged-in user's creds, and they may not have permissions to view users in other establishments. This was causing errors when requesting cell moves for some prisoners. The fix is to ignore the failed lookups as we only care about prisoners in the same establishment anyway.


GurnankCheema (approved):
LGTM

Mjwillis:
The new api will allow you to filter by the offender prison and only active NAs - so a lot less code for you to write !

You won't have to check effective dates nor the prison - you also won't have to call to lookup an prisoner as all the info will be there you need

jimbali:
Sounds great. Will it take into account the current location of the non-association, even if the NA was raised against a previous booking and so has a location of `OUTSIDE`?

PR #2529, opened 2023-08-14
NN-5364 NN-5366: get adjudication data from AdjudicationsAPI instead of PrisonAPI

neilmendum:
https://dsdmoj.atlassian.net/browse/NN-5364
https://dsdmoj.atlassian.net/browse/NN-5366


backend/api/prisonApi.ts
```diff
@@ -280,32 +280,6 @@ export const prisonApiFactory = (client) => {
```
neilmendum, line 280: I didn't update `getAdjudicationFindingTypes` (one line above) as that is not something we can source from AdjudicationsAPI

Mjwillis: But we will need to - I think its an enum - probably there is a need for exposing reference data from the API

backend/controllers/prisonerProfile/prisonerQuickLook.ts
```diff
@@ -130,7 +130,7 @@ export default ({
         incentivesApi.getIepSummaryForBooking(context, bookingId),
         prisonApi.getPositiveCaseNotes(context, bookingId, dateThreeMonthsAgo, today),
         prisonApi.getNegativeCaseNotes(context, bookingId, dateThreeMonthsAgo, today),
-        prisonApi.getAdjudicationsForBooking(context, bookingId),
+        adjudicationsApi.getAdjudicationsForBooking(context, bookingId),
```
Mjwillis, line 133: We have a future problem with this endpoint.  We don't store booking Id

Therefore we may need to either :
- (a) reinstate booking ID in the model (yuk!)
- (b) use the offender No. instead with maybe a cutoff date

Mjwillis (approved):
LGTM

PR #2553, opened 2023-09-12
MAP-134 move temporary cell transfer call new cellAllocationApi

GurnankCheema:
**DO NOT MERGE AS TEAM ARE HAVING A RETHINK ABOUT THIS FUNCTIONALITY 12/9/23**

There is a new cellAllocationApi which will take over some functionality from prisonApi.

The code change in DPS is to use the new cellAllocationApi's move-to-cell-swap endpoint rather than prisonApis's. The new api will in fact behave as a proxy as the request will be forwarded on to prisonApi.  The only difference being the requirement for system token to include the ROLE_MAINTAIN_CELL_MOVEMENTS role.


backend/controllers/cellMove/confirmCellMove.ts
```diff
@@ -85,8 +85,12 @@ export default ({ prisonApi, whereaboutsApi }) => {
...
   }
 
-  const makeCSwap = async (res, { bookingId, agencyId, offenderNo }) => {
-    await prisonApi.moveToCellSwap(res.locals, { bookingId })
+  const makeCSwap = async (req, res, { bookingId, agencyId, offenderNo }) => {
+    const { username } = req.session.userDetails
+
+    const systemContext = await systemOauthClient.getClientCredentialsTokens(username)
+
+    await cellAllocationApi.moveToCellSwap(systemContext, { bookingId })
```
GurnankCheema, line 93: This is the same as the call that was being made to prisonApi apart from the use of systemContext rather than user context

PR #2554, opened 2023-09-12
NON-208: Link to prisoner non-associations

xoen:
For non-associations private beta prisons (private beta hasn't started yet) the quicklook profile will show the number of open non-associations (if the prisoner is in a prison) and a link to the prisoner non-associations page on the Non-associations service (not live yet).

Screenshot 2023-09-12 at 10 44 47

- The `NON_ASSOCIATIONS_PRISONS` environment variable determines in which private beta prisons to show this section.
  - This is the usual comma separated list of prison IDs
  - It's currently empty for `preprod`/`prod`so this section will not be shown in these environment. 
  - In `dev` is enabled for Moorland/MDI
- The non-associations are retrieved using the currently work-in-progress Non-associations API using the system token
  - The DPS client [needs the `READ_NON_ASSOCIATIONS` role](https://non-associations-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/Legacy%20non-associations-details/getDetailsFromPrisonApiByPrisonerNumber)
- There was a `NonAssociationsApi` client with a `getNonAssociations()` method, however this is using the legacy endpoint (which is getting data from Prison API/NOMIS)
  - I've renamed that method and the corresponding return type to `getNonAssociationsLegacy()`/`OffenderNonAssociationLegacy` so that it is clear this is getting data from a different endpoint
  - I've added a `getNonAssociations()` method for the new endpoint - but this method can't be used until migration/sync is completed and Non-associations API is the source of truth.


backend/tests/prisonerQuickLook.test.ts
```diff
@@ -90,6 +136,8 @@ describe('prisoner profile quick look', () => {
 
     logError = jest.fn()
 
+    config.apis.nonAssociations.ui_url = nonAssociationsUrl
```
ushkarev, line 139: does this potentially need a clean-up `afterEach/All` step for other tests? doesn't appear matter now for the current set of tests, but there's no undo is there?

helm_deploy/values-dev.yaml
```diff
@@ -27,7 +27,9 @@ generic-service:
    CHECK_MY_DIARY_URL: https://check-my-diary-dev.prison.service.justice.gov.uk?fromDPS=true
    INCENTIVES_URL: https://incentives-ui-dev.hmpps.service.justice.gov.uk
    INCENTIVES_API_ENDPOINT_URL: https://incentives-api-dev.hmpps.service.justice.gov.uk
+   NON_ASSOCIATIONS_PRISONS: "MDI"
```
ushkarev, line 30: can we have 1 or 2 more? so that it's easier to test journeys with the "in other establishments" section on our site

ushkarev (approved):

helm_deploy/values-dev.yaml
xoen: Good point, I added a few more in this commit: https://github.com/ministryofjustice/digital-prison-services/pull/2554/commits/5d3488ac28779987b7c77b9062177890d701b46d

backend/tests/prisonerQuickLook.test.ts
xoen: Didn't see any similar cleanup blocks in the rest of the file, so I think it's OK, at least until we see it's actually leaking state and causing problems.

PR #2558, opened 2023-09-19
ESWE-646_dps_home_page_tile

monskapinga:
Create a ‘Learning and work progress’ tile on the main DPS homepage.

- When a user has the appropriate permission the Learning and work progress tile is visible.
- When the user has the appropriate permission, when they click on the tile, they will be taken to https://ciag-induction-dev.hmpps.service.justice.gov.uk/
- If the user does not have the appropriate permission they will not see the tile.


steverendell (approved):
lgtm

PR #2573, opened 2023-09-29
add cache busting for css files

whitfield-mj:


davidmiddletonmoj (approved):

Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN npm prune --production
 FROM node:16-bullseye-slim
 LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
 
-ENV CACHE_BREAKER ${GIT_REF}
+ENV GIT_REF ${GIT_REF:-xxxxxxxxxxxxxxxxxxx}
```
petergphillips, line 24: env variables don't need a default - see https://github.com/ministryofjustice/hmpps-template-typescript/pull/233/files
also better to use new `ENV MY_VAR=my_value` syntax - see _Alternative syntax_ section in https://docs.docker.com/engine/reference/builder/#env

backend/config.ts
```diff
@@ -38,6 +38,7 @@ export const app = {
     enabledPrisons: process.env.HOMEPAGE_REDIRECT_ENABLED_PRISONS || '',
     scheduleRedirectForPrisons: process.env.HOMEPAGE_SCHEDULE_REDIRECT_FOR_PRISONS || '',
   },
+  gitRef: process.env.GIT_REF || '',
```
petergphillips, line 41: should really be `requiredInProduction` - see https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/server/config.ts#L35

petergphillips (approved):
LGTM

Dockerfile
whitfield-mj: Ha, I didn't have one but included it because it was in the template!

backend/config.ts
whitfield-mj: Yes, DPS does not have a lot of things that the template does

PR #2591, opened 2023-10-17
Map 288 dps create reception moves screen

GurnankCheema:
This change is to create a new /prisoner/:prisonNumber/reception-move/consider-risks-reception page. 
It's the first of 3 new pages. 
cypress tests to be done when all pages in complete journey is available

Partial view of page
Screenshot 2023-10-17 at 08 19 38


views/partials/layout.njk
```diff
@@ -1,4 +1,14 @@
...
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+{% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/select/macro.njk" import govukSelect %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
```
jimbali, line 11: Any particular reason for putting these in the layout rather than in the individual templates?

backend/controllers/receptionMove/considerRisksReception.ts
```diff
@@ -0,0 +1,157 @@
...
+        type: nonAssociation.typeDescription,
+        selectedOffenderKey: `${formatName(firstName, lastName)} is`,
+        selectedOffenderRole: nonAssociation.reasonDescription,
+        otherOffenderKey: `${formatName(
+          nonAssociation.offenderNonAssociation.firstName,
+          nonAssociation.offenderNonAssociation.lastName
+        )} is`,
+        otherOffenderRole: nonAssociation.offenderNonAssociation.reasonDescription,
+        comment: nonAssociation.comments || 'Not entered',
+      }))
```
jimbali, line 112: If this is mostly copied from `backend/controllers/cellMove/viewNonAssociations.ts` then maybe it would be worth extracting it into a helper function?

jimbali (approved):
lgtm, just added a couple of comments

views/partials/layout.njk
GurnankCheema: Putting them in layout would mean in the future (and we could refactor current templates too) won't need to import them again. This keeps templates less cluttered.

PR #2606, opened 2023-10-25
MAP-290 fix logic that determines reception capacity

GurnankCheema:
corrected logic to display the reception-full page.
prisonApi.getReceptionsWithCapacity returns an empty array if there is no capacity so previous logic was attempting to destruct a null object


davidthomas22 (approved):
LGTM

PR #2613, opened 2023-11-01
MAP-289 added confirm reception move page

GurnankCheema:
Add the confirmation page in the move prisoner to reception feature. 
This PR does not include e2e tests. They will be done when all pages in feature ready

Screenshot 2023-11-01 at 16 18 26

Screenshot 2023-11-01 at 16 18 36


davidthomas22 (approved):
LGTM

PR #2634, opened 2023-11-13
SDIT-1242 LocationResource

steverendell:
Switch prison-api call to use a system token


backend/controllers/cellMove/cellMoveConfirmation.ts
```diff
@@ -1,6 +1,6 @@
 import { formatName, putLastNameFirst } from '../../utils'
 
-export default ({ prisonApi, logError }) =>
+export default ({ prisonApi }) =>
```
steverendell, line 3: logError seems to be a redundant hangover

petergphillips (approved):
LGTM

petergphillips (approved):
LGTM

PR #2695, opened 2024-01-31
MAP-664 bugfix

GurnankCheema:
1. content change on reception-move/consider-risks-reception page. If only one prisoner in reception, should show ‘1 person in reception’. Otherwise should display   ‘ x people in reception’ 

 2. when moving prisoner from cell to reception then back from reception to cell again, the app currently persists the radio selection button selected and the comments entered in the  confirmation pages. These should not be carried over. i.e cell-move/confirm-cell-move should not present any selections or comments that were set in /reception-move/confirm-reception-move

### DIFF
--- a/backend/controllers/receptionMove/confirmReceptionMove.ts
+++ b/backend/controllers/receptionMove/confirmReceptionMove.ts
@@ -98,6 +98,9 @@ export default ({ prisonApi, whereaboutsApi }) => {
         commentText: comment,
       })
 
+      // just to remove formValues out of flash to prevent being consumed in /cell-move/confirm-cell-move
+      req.flash('formValues')
+
       return res.redirect(`/prisoner/${offenderNo}/reception-move/confirmation`)
     } catch (error) {
       logger.error(`Error moving ${offenderNo} to reception`)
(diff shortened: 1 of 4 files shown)


backend/controllers/receptionMove/confirmReceptionMove.ts
```diff
@@ -98,6 +98,9 @@ export default ({ prisonApi, whereaboutsApi }) => {
         commentText: comment,
       })
```
GurnankCheema, line 100: This will flush Flash of any form values (radios selected, text boxes completed) so they can't be accessed later

backend/tests/receptionMove/considerRisksReceptionController.test.ts
```diff
@@ -164,7 +164,36 @@ describe('Consider risks reception', () => {
...
+
+          alerts: [],
+        },
+      ])
+      await controller.view(req, res)
+
+      expect(res.render).toHaveBeenCalledWith(
+        'receptionMoves/considerRisksReception.njk',
+        expect.objectContaining({
+          inReceptionCount: '1 person in reception',
```
GurnankCheema, line 182: line 182 is the crucial item of this test

jimbali (approved):

PR #2728, opened 2024-04-22
SDIT-1330: ⚡️ Use prisoner search to get prisoner cell location

petergphillips:


backend/controllers/appointments/viewAppointments.ts
```diff
@@ -57,69 +57,72 @@ export default ({ systemOauthClient, prisonApi, whereaboutsApi, logError }) =>
...
-              appointment.lastName
-            )}'s appointment">View details </a>`,
-            classes: 'govuk-!-display-none-print',
+    const appointmentsOfType = appointments.filter((appointment) =>
+      type ? appointment.appointmentTypeCode === type : true
+    )
+
+    const prisonerDetailsForOffenderNumbers: Array<PrisonerSearchResult> = await offenderSearchApi.getPrisonersDetails(
+      res.locals,
+      Array.from(new Set(appointmentsOfType.map((appointment) => appointment.offenderNo)))
```
petergphillips, line 66: remove duplicate prisoners by converting to set and then back to array

backend/controllers/appointments/viewAppointments.ts
```diff
@@ -57,69 +57,72 @@ export default ({ systemOauthClient, prisonApi, whereaboutsApi, logError }) =>
...
+    )
+    const cellLocationMap = new Map(prisonerDetailsForOffenderNumbers.map((i) => [i.prisonerNumber, i.cellLocation]))
```
petergphillips, line 68: construct map of prisoner to cell location

backend/tests/viewAppointmentsController.test.ts
```diff
@@ -325,7 +333,7 @@ describe('View appointments', () => {
       })
       expect(whereaboutsApi.getVideoLinkAppointments).toHaveBeenCalledWith(res.locals, [3, 4, 5, 6])
       expect(whereaboutsApi.getAgencyGroupLocationPrefix).toHaveBeenCalledWith(res.locals, 'MDI', 'H 1')
-      expect(prisonApi.getDetails).toHaveBeenCalledTimes(6)
+      expect(offenderSearchApi.getPrisonersDetails).toHaveBeenLastCalledWith(res.locals, ['ABC123', 'ABC456', 'ABC789'])
```
petergphillips, line 336: notice that we only get one call and also that we don't get duplicate prisoner numbers.  Previously we would call Prison API multiple times even for the same prisoner!

backend/tests/viewAppointmentsController.test.ts
```diff
@@ -383,7 +391,7 @@ describe('View appointments', () => {
                 attributes: { 'data-sort-value': 'FOUR' },
                 html: '<a href="/prisoner/ABC456" class="govuk-link">Four, Offender - ABC456</a>',
               },
-              { text: undefined },
+              { text: '2-1-1' },
```
petergphillips, line 394: since we've called at the start and then read from the map we have results for this prisoner.  Previous mocking only mocked the first three calls, hence last 3 were undefined.

integration-tests/integration/appointments/viewAppointments.cy.js
```diff
@@ -95,18 +95,20 @@ context('A user can view list of appointments', () => {
...
-    })
-    cy.task('stubPrisonerFullDetail', {
-      prisonerDetail: { assignedLivingUnit: { description: '3-1-1' } },
-      offenderNo: 'ABC789',
-    })
+    cy.task('stubPrisonerSearchDetails', [
+      {
+        cellLocation: '1-1-1',
+        prisonerNumber: 'ABC123',
+      },
```
mikehalmamoj, line 102: Is there a test where a cell location contains more than one prisoner?

mikehalmamoj (approved):

PR #2729, opened 2024-04-23
SDIT-1330: ⚡️ Don't call prisoner search if no prisoners to search for

petergphillips:


backend/controllers/appointments/viewAppointments.ts
```diff
@@ -1,6 +1,6 @@
 import moment from 'moment'
 import { Response } from 'express'
-import { getTime, properCaseName, getCurrentPeriod, formatName } from '../../utils'
+import { formatName, getCurrentPeriod, getTime, properCaseName } from '../../utils'
```
petergphillips, line 3: not quite sure why these were reordered 🤷, they now appear to be alphabetical

mikehalmamoj (approved):

PR #2746, opened 2024-06-24
Map 1287 dps get location data from new locations inside prison api 1

GurnankCheema:
use the new locationsInsidePrisonApi so that endpoint  '/manage-prisoner-whereabouts/select-residential-location' calls locationsInsidePrisonApi.searchGroups  rather than whereaboutsApi.searchGroups


GurnankCheema:
replaced by another branch

PR #2751, opened 2024-07-16
MAP-1409 get location-prefix from new locations api 

GurnankCheema:
get location-prefix from new locations-inside-prison-api rather than whereabouts api. 
The response to both is exactly the same and has the form of MDI-1-


Mjwillis (approved):
Lgtm

PR #2757, opened 2024-08-01
MAP-1287 now getting residential location groups from locationsApi

GurnankCheema:
moving away from whereaboutsApi to locationsInsidePrisonApi to get internal location groups.

Change applied to viewAppointments.ts so that it gets location groups from locationsInsidePrisonApi

Page affected is /view-all-appointments and component is the Residential Location select box

### DIFF
--- a/backend/controllers/appointments/viewAppointments.ts
+++ b/backend/controllers/appointments/viewAppointments.ts
@@ -54,7 +54,7 @@ export default ({ systemOauthClient, prisonApi, offenderSearchApi, whereaboutsAp
         locationId,
         offenderLocationPrefix,
       }),
-      whereaboutsApi.searchGroups(res.locals, agencyId),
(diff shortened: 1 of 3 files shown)


backend/controllers/appointments/viewAppointments.ts
```diff
@@ -54,7 +54,7 @@ export default ({ systemOauthClient, prisonApi, offenderSearchApi, whereaboutsAp
         locationId,
         offenderLocationPrefix,
       }),
-      whereaboutsApi.searchGroups(res.locals, agencyId),
+      locationsInsidePrisonApi.getSearchGroups(systemContext, agencyId),
     ])
```
GurnankCheema, line 58: L57: the key change

Mjwillis (approved):
lgtm

PR #2770, opened 2024-09-02
MICN-113 use filtered case note type and sub type endpoint

bill-moj:
this PR corresponds to the same changes in prisoner profile: https://github.com/ministryofjustice/hmpps-prisoner-profile/pull/706

- replace GET /case-notes/types-for-user call with GET /case-notes/types using the correct filters. GET /case-notes/type-for-user endpoint will be deprecated.
- set query params of what kind of CaseNoteSubTypes to be included for GET /case-notes/types calls (namely, selectable by DPS_USER or by ALL; include INACTIVE and/or RESTRICTED subtypes), based on the user's roles.

### DIFF
--- a/backend/api/caseNotesApi.ts
+++ b/backend/api/caseNotesApi.ts
@@ -57,8 +57,15 @@ export const caseNotesApiFactory = (client) => {
   const amendCaseNote = (context, offenderNo, caseNoteId, data?) =>
     client.put(context, `/case-notes/${offenderNo}/${caseNoteId}`, data).then(processResponse(context))
 
-  const getCaseNoteTypes = (context) => get(context, '/case-notes/types')
-  const myCaseNoteTypes = (context) => get(context, '/case-notes/types-for-user')
+  const getCaseNoteTypes = (context) =>
+    get(context, '/case-notes/types?selectableBy=ALL&includeInactive=true&includeRestricted=true')
+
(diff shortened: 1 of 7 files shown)


brightonsbox (approved):
LGTM

backend/api/caseNotesApi.ts
```diff
@@ -57,8 +57,15 @@ export const caseNotesApiFactory = (client) => {
...
-  const myCaseNoteTypes = (context) => get(context, '/case-notes/types-for-user')
+  const getCaseNoteTypes = (context) =>
+    get(context, '/case-notes/types?selectableBy=ALL&includeInactive=true&includeRestricted=true')
+
+  const myCaseNoteTypes = (context) => {
+    if (context?.userRoles?.find((role) => ['POM', 'ADD_SENSITIVE_CASE_NOTES'].includes(role.roleCode))) {
+      return get(context, '/case-notes/types?selectableBy=DPS_USER&includeInactive=false&includeRestricted=true')
+    }
+    return get(context, '/case-notes/types?selectableBy=DPS_USER&includeInactive=false&includeRestricted=false')
+  }
```
brightonsbox, line 68: It looks like we're a little light on testing (appreciate it probably already was) - would perhaps expect to see the different calls being exercised in tests depending on user roles

PR #2783, opened 2024-10-01
MICN-157 change all case notes api call to use client token

bill-moj:
This PR updates case-notes related API calls to follow the updated specs of offender-case-notes.

1. added a `getContextWithClientTokenAndRoles` function in prisonProfileContext, which always return a context with client token. this context is used by all case notes api calls.
2. for `getCaseNote` calls, a custom header `CaseloadId` is added.
3. for `getCaseNotes` API calls, an additional query param `includeSensitive=true` is added if the users has certain specified role(s).

### DIFF
--- a/backend/controllers/controller.ts
+++ b/backend/controllers/controller.ts
@@ -120,8 +120,6 @@ export const factory = ({
     const prisonerDetails = await prisonApi.getDetails(res.locals, offenderNumber)
     if (prisonerDetails?.agencyId) {
       contextProperties.setCustomRequestHeaders(context, { CaseloadId: prisonerDetails.agencyId })
-    } else if (req.session?.userDetails?.activeCaseLoadId) {
-      contextProperties.setCustomRequestHeaders(context, { CaseloadId: req.session.userDetails.activeCaseLoadId })
     }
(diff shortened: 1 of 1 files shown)


backend/controllers/controller.ts
```diff
@@ -105,7 +108,24 @@ export const factory = ({
...
+      res,
+      req,
+      oauthApi: null,
+      systemOauthClient: null,
+      restrictedPatientApi: null,
+    })
+    try {
+      const prisonerDetails = await prisonApi.getDetails(res.locals, offenderNumber)
+      contextProperties.setCustomRequestHeaders(context, { CaseloadId: prisonerDetails.agencyId })
+    } catch (e) {
```
iterative-it, line 122: What could fail here that would cause this exception? Could do prisonerDetails?.agencyId to get rid of the try catch perhaps?

bill-moj: i added the try catch because i was afraid of potential exception from `prisonApi.getDetails`
but then the int tests are passing now so it should be okay to go without this try catch.

timoj58 (approved):

PR #2803, opened 2024-12-01
Map old Prison API locations to new Locations for Video Link Bookings

smcveigh941:
The whereabouts API works using NOMIS location ids.
The BVLS API works using LocationsAPI location ids.

Up until now, we have assumed that the locationPrefix of a nomis location is equal to the business key of the locations returned from Locations API. However, this assumption is not always the case, namely at HMP Thameside.

Nomis location ids (fetched from Whereabouts API) must be mapped to DPS location keys using the NomisMapping tool provided by Syscon, for use when communicating with BVLS API and vice versa


timharrison-moj (commented):
LGTM. This is to be tested in PREPROD with the known good and "bad" locations already identified at Thameside.

matt-crowson-moj (approved):
LGTM

PR #2807, opened 2024-12-13
BAVL-440: Remove the feature toggle for BVLS

smcveigh941:


backend/tests/appointmentDetails.test.ts
```diff
@@ -68,6 +68,18 @@ describe('appointment details', () => {
     // @ts-expect-error ts-migrate(2339) FIXME: Property 'getAppointment' does not exist on type '... Remove this comment to see the full error message
     whereaboutsApi.getAppointment = jest.fn().mockResolvedValue(testAppointment)
```
timharrison-moj, line 69: Should it be calling whereaboutsApi to get an appointment here? Suppose this is not a VLB, so maybe yes

timharrison-moj (approved):
Nice. LGTM

PR #2815, opened 2025-01-08
BAVL-553 excluding video link court hearing appointment types from bulk appointment bookings. Court hearing video appointments should be managed by BVLS only.

matt-crowson-moj:
During the decommissioning of the legacy BVLS code, a gap was noticed where you can bulk book court hearing video links entirely bypassing the BVLS service (old and new versions).  These go straight into NOMIS and BVLS has no knowledge of these appointments.

Note, as part of this change I have also excluded probation meetings as the same applies for these also.

This should not be allowed.  All of these appointment types should go through the BVLS service.

### DIFF
--- a/views/bulkAppointmentsAddDetails.njk
+++ b/views/bulkAppointmentsAddDetails.njk
@@ -101,7 +101,7 @@
                         label: {
                             text: "Appointment type"
                         },
-                        items: appointmentTypes | addDefaultSelectedVale('Select') | setSelected(appointmentType),
+                        items: appointmentTypes | filterNot('value', ['VLB', 'VLPM']) | addDefaultSelectedVale('Select') | setSelected(appointmentType),
                         errorMessage: errors | findError('appointment-type')
                     }) }}
 
--- a/backend/utils.test.ts
+++ b/backend/utils.test.ts
@@ -25,6 +25,7 @@ import {
   properCaseName,
   putLastNameFirst,
   stringWithAbbreviationsProcessor,
+  filterNot,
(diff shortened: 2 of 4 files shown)


views/bulkAppointmentsAddDetails.njk
```diff
@@ -101,7 +101,7 @@
                         label: {
                             text: "Appointment type"
                         },
-                        items: appointmentTypes | addDefaultSelectedVale('Select') | setSelected(appointmentType),
+                        items: appointmentTypes | filterNot('value', ['VLB']) | addDefaultSelectedVale('Select') | setSelected(appointmentType),
```
matt-crowson-moj, line 104: VLB could be pulled out into a constant/var for clarity?

views/bulkAppointmentsAddDetails.njk
```diff
@@ -101,7 +101,7 @@
                         label: {
                             text: "Appointment type"
                         },
-                        items: appointmentTypes | addDefaultSelectedVale('Select') | setSelected(appointmentType),
+                        items: appointmentTypes | filterNot('value', ['VLB', 'VLPM']) | addDefaultSelectedVale('Select') | setSelected(appointmentType),
```
matt-crowson-moj, line 104: Should these codes be pulling out into variables with better names?

timharrison-moj: I think its fine as is. This code is (hopefully) short-lived.

timharrison-moj (approved):
LGTM

PR #2817, opened 2025-01-09
BAVL-549 first set of changes towards removing legacy BVLS specific code from DPS, this is now handled via a redirect to the prisoner profile add appointment.

matt-crowson-moj:
This is the first set of changes towards removal of legacy code/logic for video link court/probation bookings.

These type of bookings are taken care of elsewhere.

The question remains how far to take this e.g. the 'add appointment' itself has not been removed entirely only the bits around court/probation bookings.

There is also a separate task to remove gov notifiy related code as this is only used for video link bookings.


backend/controllers/appointments/addAppointment.ts
matt-crowson-moj, line 1: So this file still exists.  Strictly speaking it shouldn't be accessed by anything as this should be handled via the prisoner profile page.

In the interim I have removed anything VLB specific.

backend/api/whereaboutsApi.ts
```diff
@@ -59,10 +59,6 @@ export const whereaboutsApiFactory = (client) => {
       `/attendances/offender/${offenderNo}/unacceptable-absences?fromDate=${fromDate}&toDate=${toDate}&page=${page}`
     )
 
-  const getCourtLocations = (context) => get(context, '/court/courts')
```
matt-crowson-moj, line 62: The endpoints (intentionally) no longer exists on whereabouts-api.

backend/controllers/appointments/addAppointment.ts
matt-crowson-moj, line 1: I have not removed 'add appointment' entirely only those bits related to VLB's.

timharrison-moj (approved):
LGTM  - all functionality that should not be accessible anyway.

PR #2818, opened 2025-01-10
BAVL-550 removing now redundant legacy BVLS email code/configuration.

matt-crowson-moj:
This code is no longer required, the code that was using this has since been removed.  The new replacement BVLS service is now handling all emails.

There will be some manual steps here to remove the k8s config values.  Note removal of values from K8s should only done once all of these changes have been deployed as far as prod (in case there is a need to rollback).


helm_deploy/digital-prison-services/values.yaml
```diff
@@ -52,21 +52,6 @@ generic-service:
...
       SESSION_COOKIE_SECRET: "SESSION_COOKIE_SECRET"
-      NOTIFY_API_KEY: "NOTIFY_API_KEY"
-      WANDSWORTH_OMU_EMAIL: "WANDSWORTH_OMU_EMAIL"
-      THAMESIDE_OMU_EMAIL: "THAMESIDE_OMU_EMAIL"
-      HEWELL_OMU_EMAIL: "HEWELL_OMU_EMAIL"
-      BERWYN_OMU_EMAIL: "BERWYN_OMU_EMAIL"
-      ELMLEY_OMU_EMAIL: "ELMLEY_OMU_EMAIL"
-      NOTTINGHAM_OMU_EMAIL: "NOTTINGHAM_OMU_EMAIL"
-      BULLINGDON_OMU_EMAIL: "BULLINGDON_OMU_EMAIL"
-      PETERBOROUGH_OMU_EMAIL: "PETERBOROUGH_OMU_EMAIL"
```
timharrison-moj, line 63: These are also configured in the k8s secrets. Probably manually? But need a note on the PR to remove them?

timharrison-moj (approved):
LGTM

helm_deploy/digital-prison-services/values.yaml
matt-crowson-moj: There is a note on the PR.  Did you mean the Jira ticket?

PR #2821, opened 2025-01-21
BAVL-572 & BAVL-573: Include location when mapping an appointment to a video link booking, and fix added by field for probation users

smcveigh941:
* Fix issue where bookings were being mismatched between Whereabouts and BVLS
* Fix issue where probation bookings would say `Added by: Court`

### DIFF
--- a/backend/controllers/appointments/viewAppointments.ts
+++ b/backend/controllers/appointments/viewAppointments.ts
@@ -40,6 +40,7 @@ export default ({
     whereaboutsApi,
     locationsInsidePrisonApi,
     bookAVideoLinkApi,
+    nomisMapping,
   }) =>
   async (req, res: Response) => {
     // @ts-expect-error ts-migrate(2554) FIXME: Expected 1 arguments, but got 0.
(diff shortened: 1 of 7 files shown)


backend/controllers/appointments/viewAppointments.ts
```diff
@@ -85,7 +93,6 @@ export default ({
       })
       .filter(Boolean)
 
-    const videoLinkAppointmentsMadeByTheCourt = videoLinkAppointments.filter((app) => app.madeByTheCourt)
```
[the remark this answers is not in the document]
timharrison-moj: I made a note on the ticket about this line, as there are no attributes called `madeByTheCourt` any more. Nice to see it gone.

backend/controllers/appointments/viewAppointments.ts
```diff
@@ -66,13 +67,20 @@ export default ({
 
     const videoLinkAppointments = await bookAVideoLinkApi.getPrisonSchedule(systemContext, agencyId, searchDate)
 
+    const locationMappings = await Promise.all(
+      [...new Set(videoLinkAppointments.map((vlb) => vlb.dpsLocationId))].map((id) =>
+        nomisMapping.getNomisLocationMappingByDpsLocationId(systemContext, id)
+      )
+    )
```
timharrison-moj, line 74: Its one call for every VLB appointment in the list? But it does get the unique key. I was thinking could cut a couple of corners and use the `prisonLocDesc` to match against the `locationDescription` - they both hold what the locations API calls the `localName`.

timharrison-moj: Might be an issue if the numbers of VLBs (and later, VLPMs) is very long. Is there a mapping POST endpoint which takes a list of IDs?

timharrison-moj: Yes, that's fine - go for it your way and let's see how it performs. If you look at Sunday 18th Jan in Birmingham - I created two appointments / VLBs that were being mixed up.

timharrison-moj (approved):
LGTM

PR #2838, opened 2025-02-13
CDPS-1197: Removing redirect exemptions flag

brightonsbox:
For a while now, there have been no `PRISONER_PROFILE_REDIRECT_EXEMPTIONS` so this config can be removed.  However in doing so, we can no longer use exemption caseload configuration in integration tests - the integration tests now show the reality of the remaining users that can still access the old profile - those with no caseloads.


backend/controllers/prisonerProfile/prisonerProfileBackLinkRedirect.ts
```diff
@@ -1,21 +1,13 @@
 import config from '../../config'
-import { isRedirectCaseLoad } from '../../utils'
```
brightonsbox, line 2: Removing `isRedirectCaseLoad` since this was based purely on exemption case loads which this PR is removing.  This means that this always redirects to the Prisoner Profile, which is why `backLink.ts` is removed above

backend/controllers/prisonerProfile/prisonerProfileRedirect.ts
```diff
@@ -11,9 +10,7 @@ export default ({ path, handler }) => {
       `Old prisoner profile disabled from: ${config.app.prisonerProfileRedirect.oldPrisonerProfileInaccessibleFrom}`
     )
 
-    const redirectEnabled = isRedirectCaseLoad(activeCaseLoadId)
-
-    if (activeCaseLoadId && redirectEnabled) {
+    if (activeCaseLoadId) {
```
brightonsbox, line 13: Without the concept of exemptions, we can see clearer that the Prisoner Profile redirect is now solely on whether the user has an activeCaseLoad set!

backend/utils.ts
```diff
@@ -344,10 +344,6 @@ export const stringWithAbbreviationsProcessor = (string: string): string => {
   return establishmentName
 }
 
-export const isRedirectCaseLoad = (activeCaseLoadId: string): boolean => {
```
brightonsbox, line 347: This was purely taking into account the exemptions list which we are getting rid of in this PR

integration-tests/integration/alerts/createAlert.cy.js
```diff
@@ -7,7 +7,7 @@ const NotFoundPage = require('../../pages/notFound')
 context('A user can add an alert', () => {
   beforeEach(() => {
     cy.task('resetAndStubTokenVerification')
-    cy.task('stubSignIn', { username: 'ITAG_USER', caseload: 'MDI', roles: ['ROLE_UPDATE_ALERT'] })
```
brightonsbox, line 10: Given there are no exemptions in the cypress config any more, the only way many endpoints can be accessed is now if the user has no caseloads

integration-tests/integration/caseNotes/case-note-amendments.cy.js
```diff
@@ -94,7 +96,8 @@ context('Case note amendments', () => {
     page.moreDetail().type('Hello, world!')
     page.save().click()
 
-    CaseNotesPage.verifyOnPage('Smith, John')
+    // Weirdly a user can create the case note but can't access the case note page...
```
brightonsbox, line 99: Uncovering some interesting facts about how users with no caseloads would have been able to interact with the old profile!

integration-tests/integration/prisonerProfile/caseNotesPage.cy.js
```diff
@@ -1,335 +1,17 @@
...
-  ],
-  eventId: -23,
-}
-
-const replicate = ({ data, times }) =>
-  Array(times)
-    .fill()
-    .map(() => data)
-
-context('A user can view prisoner case notes', () => {
```
brightonsbox, line 40: This is not possible now

integration-tests/integration/prisonerProfile/prisonerWorkAndSkills.cy.js
```diff
@@ -559,9 +549,8 @@ context('Prisoner Work and Skills', () => {
           cy.get($workScheduleLink).contains('View 7 day schedule')
         })
 
-        cy.get('[data-test="work-unacceptableAbsence"]').contains('Last 30 days')
-        cy.get('[data-test="work-unacceptableAbsence"]').contains('4')
-        cy.get('[data-test="work-absencesLink"]').contains('View unacceptable absences for the last 6 months')
+        cy.get('[data-test="work-unacceptableAbsence"]').should('not.exist')
+        cy.get('[data-test="work-absencesLink"]').should('not.exist')
```
brightonsbox, line 553: Some bits of the work and skills were conditional on a list of 'accelerator' caseloads containing the user's active caseload - now the only way the user can access this is if they have NO caseload, so these bits are no longer visible at all

views/prisonerProfile/prisonerProfessionalContacts/prisonerProfessionalContacts.njk
```diff
@@ -76,9 +76,5 @@
     {% endif %}
   {% endfor %}
 
-  {% if useNewProfile %}
```
brightonsbox, line 79: This was always true with no exemptions

scottrowley (approved):

PR #2857, opened 2025-02-24
CDPS-1252: Updating NodeJS to 22.14

brightonsbox:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:20.9-bullseye-slim as builder
+FROM node:22.9-bullseye-slim as builder
```
davidatmoj, line 1: Should this be 22.14-bookworm-slim ?
And below too (bookworm)

davidatmoj (approved):

PR #2895, opened 2025-04-30
BAVL-813 removing the master VLPM appointments feature toggle, now safe to remove.

matt-crowson-moj:
This BVLS feature toggle is now safe to be removed.  The supported appointments types for VLB and VLPM are now well and truly live.


backend/config.ts
```diff
@@ -38,11 +38,7 @@ export const app = {
     scheduleRedirectForPrisons: process.env.HOMEPAGE_SCHEDULE_REDIRECT_FOR_PRISONS || '',
     exemptions: process.env.HOMEPAGE_REDIRECT_EXEMPTIONS || '',
   },
-  bvlsMasteredVlpmFeatureToggleEnabled: process.env.BVLS_MASTERED_VLPM_FEATURE_TOGGLE_ENABLED === 'true',
-  bvlsMasteredAppointmentTypes: [
-    'VLB',
-    process.env.BVLS_MASTERED_VLPM_FEATURE_TOGGLE_ENABLED === 'true' ? 'VLPM' : undefined,
-  ].filter(Boolean),
+  bvlsMasteredAppointmentTypes: ['VLB', 'VLPM'].filter(Boolean),
   gitRef: process.env.GIT_REF || '',
```
timharrison-moj, line 42: Does it need the filter(Boolean) any more?

timharrison-moj (approved):
LGTM. Small question on filter needed.

backend/config.ts
matt-crowson-moj: This isn't the toggle.  This is the supported appointment types.

timharrison-moj: Yup, I realise that. I see what its doing on 2nd look - the `.filter(Boolean)` part is there so it can be used in a certain way. All good.

PR #2937, opened 2025-06-02
BAVL-901 small content changes for the BVLS court and probation appointments containing notes.

matt-crowson-moj:
Small content change for BLVS appointments to show 'None entered' instead of 'Not entered' for notes when none present for consistency across services.


timharrison-moj (approved):
LGTM

PR #2958, opened 2025-07-02
BAVL-949 removing old comments code that has been sitting behind the public and private notes feature toggle.

matt-crowson-moj:
Removing no longer needed (old) BVLS code now that we are happy the feature toggled code is running fine.

I have run this locally as well and appears to be working as expected.

Screenshots showing 2 BVLS bookings and one not.

dps-court

dps-probation

dps-canteen


sameer-moj (approved):
LGTM

PR #2993, opened 2025-09-16
BAVL-1081 removal of feature guest PIN/link toggle

matt-crowson-moj:
Guest PIN / link functionality has now been live long enough to warrant the removal of the feature toggle code that sits around it.


timharrison-moj (approved):
LGTM

PR #3005, opened 2025-10-02
CDPS-1617: Use shared items lib for alert flags

brightonsbox:


backend/controllers/search/prisonerSearch.ts
```diff
@@ -127,10 +127,7 @@ export default ({
           assignedLivingUnitDesc: formatLocation(prisoner.assignedLivingUnitDesc),
           name: putLastNameFirst(prisoner.firstName, prisoner.lastName),
           alerts: alertFlagLabels.filter((alertFlag) =>
-            alertFlag.alertCodes.some(
-              (alert) =>
-                prisoner.alertsDetails && prisoner.alertsDetails.includes(alert) && profileAlertCodes.includes(alert)
```
brightonsbox, line 132: profileAlertCodes match up exactly with the codes in the library's alertFlagLabels so we can remove this array and the checks

PR #3006, opened 2025-10-03
CDPS-1617: Update Docker file, move to alpine, and fix asset cache busting

brightonsbox:


Dockerfile
```diff
@@ -1,59 +1,78 @@
...
-RUN npm prune --production
+# Define env variables for runtime health / info
+ENV BUILD_NUMBER=${BUILD_NUMBER}
+ENV GIT_REF=${GIT_REF}
+ENV GIT_BRANCH=${GIT_BRANCH}
 
-FROM node:22.14-bookworm-slim
-LABEL maintainer="HMPPS Digital Studio <info@digital.justice.gov.uk>"
+# Stage: build assets
+FROM base AS build
```
brightonsbox, line 34: I *think* that this wasn't working because `ARG GIT_REF` wasn't defined again in this second stage

ushkarev (approved):

PR #3148, opened 2026-07-01
MAPB-579 Migrate to non-residential service-type locations endpoint

Mjwillis:
## What
Migrate off the deprecated locations-inside-prison-api endpoint `GET /locations/prison/{prisonId}/non-residential-usage-type/{usageType}` to its replacement `GET /locations/non-residential/prison/{prisonId}/service/{serviceType}` (`serviceType=APPOINTMENT`).

## Why
The usage-type endpoint is marked `deprecated: true`. This service was one of three still calling it. The replacement is on the API's main (MAP-2930 / PR #477).

## Notes
- `usageType=APPOINTMENT` maps directly to `serviceType=APPOINTMENT` (per the API's `ServiceType` enum).
- Same auth role (`VIEW_LOCATIONS`), same query params (`formatLocalName`/`sortByLocalName`) and identical `Location` response shape, so no call-site or response-mapping changes were needed.
- Client path, unit test and WireMock/nock stub updated; unit tests pass.


davidthomas22 (approved):
LGTM

